PR #11, opened 2023-07-11
Example script that takes snapshots

dguerizec:
Note: This PR is based on https://github.com/livekit/client-sdk-python/pull/10


theomonnom:
Hi, thanks but we may have to create a server sdk before

PR #39, opened 2023-08-16
feat: better build system

theomonnom:
- windows arm64 is ignored
- musllinux is also ignored

download prebuilt ffi library from the Rust repo at build time
also build platform wheel to reduce binary size


davidzhao (approved):
lgtm!

how does it handle the upload to pypi part?

setup.py
```diff
@@ -12,40 +12,85 @@
...
+        'bdist_wheel': bdist_wheel,
+        'build_py': BuildPyCommand,
+    },
     classifiers=[
-        "Development Status :: 3 - Alpha",
         "Intended Audience :: Developers",
         "License :: OSI Approved :: Apache Software License",
+        "Topic :: Multimedia :: Sound/Audio",
+        "Topic :: Multimedia :: Video",
+        "Topic :: Scientific/Engineering :: Artificial Intelligence",
```
davidzhao, line 77: 🥇

setup.py
```diff
@@ -12,40 +12,85 @@
...
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3 :: Only",
     ],
-
-    keywords="webrtc, livekit",
+    keywords=["webrtc", "realtime", "audio", "video", "livekit"],
     license="Apache-2.0",
     packages=["livekit"],
-    python_requires=">=3.7, <4",
+    python_requires=">=3.7.0",
```
davidzhao, line 88: do we really need 3.7? or is 3.6 ok too?

davidzhao (approved):
🚀

setup.py
theomonnom: We use dataclass, I just saw this is from Python 3.7: https://peps.python.org/pep-0557/

theomonnom:
The upload part will be inside another PR

PR #41, opened 2023-08-21
Allow asynchronous event callbacks

dguerizec:
In the current code, event callbacks cannot be async.
This PR is meant to be able to define async event callbacks.

### DIFF
--- a/livekit/audio_stream.py
+++ b/livekit/audio_stream.py
@@ -14,7 +14,7 @@
 
 from weakref import WeakValueDictionary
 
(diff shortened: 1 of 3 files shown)


theomonnom:
Are we sure the events are still ordered with AsyncIOEventEmitter?

dguerizec:
Mmmm, probably not.
I'll check my use case again to see if I absolutely need this.

dguerizec:
Closing since I realized it's not necessary.

PR #48, opened 2023-08-30
e2ee support.

cloudwebrtc:
- [x] python wrappers for e2ee
- [x] examples

examples:
publish hue video with e2ee 
[e2ee_hue.py](https://github.com/livekit/client-sdk-python/blob/duan/e2ee/examples/e2ee_hue.py)

publish wave with e2ee
 [e2ee_wave.py](https://github.com/livekit/client-sdk-python/blob/duan/e2ee/examples/e2ee_wave.py)

publish and subscribe
[e2ee_wave_recv.py](https://github.com/livekit/client-sdk-python/blob/duan/e2ee/examples/e2ee_wave_recv.py)
[e2ee_hue_recv.py](https://github.com/livekit/client-sdk-python/blob/duan/e2ee/examples/e2ee_hue_recv.py)


livekit/e2ee.py
```diff
@@ -0,0 +1,195 @@
...
+
+from ._ffi_client import FfiHandle, ffi_client
+from ._proto import e2ee_pb2
+from ._proto import ffi_pb2 as proto_ffi
+
+DEFAULT_RATCHET_SALT = b"LKFrameEncryptionKey"
+DEFAULT_MAGIC_BYTES = b"LK-ROCKS"
+DEFAULT_RATCHET_WINDOW_SIZE = 16
+
+class EncryptionType(enum.Enum):
```
davidzhao, line 28: It's great to have these python classes! do these map to FFI types?

livekit/room.py
```diff
@@ -34,6 +36,7 @@
 class RoomOptions:
     auto_subscribe: bool = True
     dynacast: bool = True
+    e2ee_options: Optional[E2EEOptions] = None
```
davidzhao, line 39: since they are all options, `_options` suffix seems redundant

```suggestion
    e2ee: Optional[E2EEOptions] = None
```

davidzhao (commented):
This is great work! 🔥 

A few comments. Will let @theomonnom give the final 👍

cloudwebrtc:
e2ee_wave_recv.py/e2ee_hue_recv.py in the PR for python e2ee  is to check that the decrypted audio/video raw data can be correctly received in e2ee mode. These two examples are not necessary. you can remove it if you want.
Receiving audio/video in e2ee mode works correctly in the rust example, so it should work in Python too

PR #51, opened 2023-09-06
fix: independent event loops & streams   

keepingitneil:
Previously, AudioStreams and VideoStreams would emit frames as an event. Here we make them async iterators. This makes working with streams a bit more ergonomic. State specific to a stream can live with the logic of frame handling. For example:

```
   async def transcribe_loop(stream: AudioStream):

        written_samples = 0
        buffer = np.zeros(WHISPER_SAMPLE_RATE * SECONDS_PER_WHISPER_INFERENCE, dtype=np.float32)
        async for frame_coro in agent_stream:
            frame = await frame_coro
            resampled = frame.remix_and_resample(WHISPER_SAMPLE_RATE, 1)
            data = np.array(resampled.data, dtype=np.float32) / 32768.0
            buffer[written_samples: written_samples + len(data)] = data
            written_samples += len(data)

            if written_samples >= WHISPER_SAMPLE_RATE * SECONDS_PER_WHISPER_INFERENCE:
                res = self.whisper.transcribe(buffer)
                written_samples = 0
                print(res)
```

### DIFF
--- a/livekit/_ffi_client.py
+++ b/livekit/_ffi_client.py
@@ -17,7 +17,8 @@ import ctypes
 import os
 import platform
 import threading
-from typing import Callable, List
+from contextlib import contextmanager
+from typing import Callable, Generic, List, TypeVar
 
 import pkg_resources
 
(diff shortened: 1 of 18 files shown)


theomonnom:
not safe to merge atm, need to ensure synchronisations are still OK

theomonnom:
sorry @keepingitneil  for stealing your PR with my changes lol

theomonnom:
Important commit is https://github.com/livekit/client-sdk-python/pull/51/commits/22137062f91c870067a90d18db028d9515acf2d6

keepingitneil:
> sorry @keepingitneil for stealing your PR with my changes lol

haha teamwork!

livekit/_ffi_client.py
```diff
@@ -68,26 +69,50 @@ def ffi_event_callback(data_ptr: ctypes.POINTER(ctypes.c_uint8),  # type: ignore
...
         self.request(req)
 
+    async def wait_for_event(self, fnc: Callable[[proto_ffi.FfiEvent], bool]) \
+            -> proto_ffi.FfiEvent:
+        queue = self.subscribe()
+        try:
+            while True:
+                event = await queue.get()
+                if fnc(event):
+                    return event
```
keepingitneil, line 96: @theomonnom when do we remove the event from the queue?

theomonnom: get is removing it

keepingitneil: Isn't this a problem then where we remove it even if fn returns False?

PR #52, opened 2023-09-06
Fixed face_landmark demo with event loop workaround

davidzhao:


davidzhao:
Will hold off due to event loop + Stream iterator change.

theomonnom:
fixed by #51

PR #56, opened 2023-09-08
Update README.md

dsa:


examples/whisper/README.md
```diff
@@ -2,24 +2,42 @@
...
+### Step 4: 
+Download a model you want to use, for example:
+```
+./models/download-ggml-model.sh tiny.en
+```
+
+### Step 5: 
+Rename the shared object library:
+```
+mv libwhisper.so libwhisper.dylib
```
theomonnom, line 30: it depends on the platform:
dylib for macos
so for linux
dll for windows

theomonnom (approved):

PR #59, opened 2023-09-08
feat: max stream capacity & fix leaks when stream ends

theomonnom:
- don't miss any event by subscribing to events before creating the stream
- make sure we always create a FfiHandle for all incoming frames before closing the task
- add stream capacity (avoid OOM if the user code is slow)


davidzhao (approved):

livekit/audio_stream.py
```diff
@@ -30,20 +39,28 @@ def __init__(self, track: Track) -> None:
...
+
+    async def _run(self):
+        while True:
+            event = await self._ffi_queue.wait_for(self._is_event)
+            audio_event = event.audio_stream_event
+
+            if audio_event.HasField('frame_received'):
+                owned_buffer_info = audio_event.frame_received.frame
+                frame = AudioFrame(owned_buffer_info)
+                self._queue.put(frame)
```
keepingitneil, line 53: I'm not sure if it makes more sense to drop already enqueued frames with the ringbuffer vs just dropping new incoming frames with a standard queue that has a capacity.

Also we should log a warning if we ever drop frames

theomonnom: my guess is that sometimes it is intended to only want the most recent frames.

keepingitneil: i don't have a super strong opinion because this is a case where things are failing. Just if we drop frames when we get them (if the queue is full) we don't need to introduce the ring buffer and can more easily log which frames are dropped.

davidzhao: for a real-time system, I think we should drop older frames and keep newer ones.

PR #60, opened 2023-09-12
fix: keep events order

theomonnom:
- Always ensure we keep the order from the FfiServer
   - So the complex logic can be handled from Rust (for synchronisation)
   - E.g: the ffi server ensure that the `LocalTrackPublished` callback is sent before the `LocalTrackPublished` event
- We know use only one FfiQueue per room (So it is possible for us to keep things synchronised)

The room musts be used inside one event_loop (specified inside the ctor)
AudioStream & VideoStream can be used on other event_loops (making it easy to have another IO thread for e.g)

### DIFF
--- a/examples/basic_room.py
+++ b/examples/basic_room.py
@@ -47,24 +47,18 @@ async def main() -> None:
                              participant: livekit.RemoteParticipant):
         logging.info("track unpublished: %s", publication.sid)
 
-    # Keep a reference to the streams, otherwise they will be disposed
-    audio_stream = None
-    video_stream = None
-
     @room.listens_to("track_subscribed")
     def on_track_subscribed(track: livekit.Track,
                             publication: livekit.RemoteTrackPublication,
                             participant: livekit.RemoteParticipant):
         logging.info("track subscribed: %s", publication.sid)
         if track.kind == livekit.TrackKind.KIND_VIDEO:
-            nonlocal video_stream
             video_stream = livekit.VideoStream(track)
             # video_stream is an async iterator that yields VideoFrame
         elif track.kind == livekit.TrackKind.KIND_AUDIO:
             print("Subscribed to an Audio Track")
-            nonlocal audio_stream
             audio_stream = livekit.AudioStream(track)
-            # audio_stream is an async iterator that yields AudioFrame 
+            # audio_stream is an async iterator that yields AudioFrame
 
     @room.listens_to("track_unsubscribed")
     def on_track_unsubscribed(track: livekit.Track,
@@ -142,10 +136,10 @@ if __name__ == "__main__":
                         logging.FileHandler("basic_room.log"), logging.StreamHandler()])
 
     loop = asyncio.get_event_loop()
-    main_task = asyncio.ensure_future(main())
+    asyncio.ensure_future(main())
     for signal in [SIGINT, SIGTERM]:
-        loop.add_signal_handler(signal, main_task.cancel)
+        loop.add_signal_handler(signal, loop.stop)
     try:
-        loop.run_until_complete(main_task)
+        loop.run_forever()
     finally:
         loop.close()
--- a/examples/basic_room.py
+++ b/examples/basic_room.py
@@ -114,21 +114,12 @@ async def main() -> None:
     def on_reconnected() -> None:
         logging.info("reconnected")
 
-    try:
-        logging.info("connecting to %s", URL)
-        await room.connect(URL, TOKEN)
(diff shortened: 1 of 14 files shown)


livekit/_utils.py
```diff
@@ -22,3 +24,85 @@ async def get(self) -> T:
             await self._event.wait()
         self._event.clear()
         return self._queue.popleft()
+
+
+class BroadcastQueue(Generic[T]):
```
theomonnom, line 29: Better naming for BroadcastQueue & BroadcastChannel.
Feels confusing rn

livekit/_utils.py
```diff
@@ -22,3 +24,90 @@ async def get(self) -> T:
...
         self._event.clear()
         return self._queue.popleft()
+
+
+class Queue(asyncio.Queue[T]):
+    def __init__(self,
+                 loop: Optional[asyncio.AbstractEventLoop] = None,
+                 maxsize: int = 0) -> None:
+        super().__init__(maxsize)
+        self._loop = loop or asyncio.get_event_loop()
```
theomonnom, line 34: Am I right to assume self._loop will be used for the asyncio.Queue?

livekit/_utils.py
davidzhao: agreed it's a bit confusing.

also a bit confusing that `subscribe` returns a queue for clients to manage.

davidzhao: if the main difference between _Queue and _Channel is thread safety? why not call it ThreadsafeBroadcastQueue?

livekit/_utils.py
davidzhao: how would it know to use self._loop?

theomonnom: this class inherits asyncio.Queue, which also use self._loop.

davidzhao: ok, I don't think it's safe to override loop in that way.. 

https://github.com/python/cpython/blob/b88d9e75f68f102aca45fa62e2b0e2e2ff46d810/Lib/asyncio/mixins.py#L9

theomonnom: right

theomonnom: it is only being overrided when not used but I guess I could try to find a better solution

theomonnom:
it's tricky, now with the changes, we can't publish tracks before running room.run() ...

examples/publish_hue.py
```diff
@@ -58,32 +58,24 @@ async def main():
...
 
     loop = asyncio.get_event_loop()
-    main_task = asyncio.ensure_future(main())
+    asyncio.ensure_future(main())
     for signal in [SIGINT, SIGTERM]:
-        loop.add_signal_handler(signal, main_task.cancel)
+        loop.add_signal_handler(signal, loop.stop)
     try:
-        loop.run_until_complete(main_task)
+        loop.run_forever()
```
davidzhao, line 79: nice!

examples/basic_room.py
```diff
@@ -6,7 +6,7 @@
 import livekit
 
 URL = 'ws://localhost:7880'
-TOKEN = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE5MDY2MTMyODgsImlzcyI6IkFQSVRzRWZpZFpqclFvWSIsIm5hbWUiOiJuYXRpdmUiLCJuYmYiOjE2NzI2MTMyODgsInN1YiI6Im5hdGl2ZSIsInZpZGVvIjp7InJvb20iOiJ0ZXN0Iiwicm9vbUFkbWluIjp0cnVlLCJyb29tQ3JlYXRlIjp0cnVlLCJyb29tSm9pbiI6dHJ1ZSwicm9vbUxpc3QiOnRydWV9fQ.uSNIangMRu8jZD5mnRYoCHjcsQWCrJXgHCs0aNIgBFY'
```
davidzhao, line 9: can't wait till we have server-sdk integrated so this isn't necessary

livekit/room.py
```diff
@@ -64,6 +70,7 @@ def name(self) -> str:
     def metadata(self) -> str:
         return self._info.metadata
 
+    @property
     def e2ee_manager(self) -> E2EEManager:
```
davidzhao, line 74: would this change the interface? let's document in changelog

davidzhao (approved):
lgtm!

livekit/room.py
theomonnom: 👍

PR #61, opened 2023-09-14
feat: add server sdk

theomonnom:


.github/workflows/build-api.yml
```diff
@@ -0,0 +1,61 @@
+name: Build API
+
+on:
+  push:
+    branches: ["main"]
+    tags: ["api-v*"]
```
davidzhao, line 6: would be good to filter only api path?

```
    paths:
      - livekit-api/**
```

.github/workflows/build-rtc.yml
```diff
@@ -1,15 +1,18 @@
-name: Build
+name: Build RTC
 
 on:
   push:
     branches: ["main"]
-    tags: ["v*"]
+    tags: ["rtc-v*"]
```
davidzhao, line 6: same as above

examples/face_landmark/face_landmark.py
```diff
@@ -69,18 +69,18 @@ def draw_landmarks_on_image(rgb_image, detection_result):
             .get_default_face_mesh_iris_connections_style())
 
 
-async def frame_loop(video_stream: livekit.VideoStream) -> None:
+async def frame_loop(video_stream: rtc.VideoStream) -> None:
     landmarker = FaceLandmarker.create_from_options(options)
     argb_frame = None
-    cv2.namedWindow('livekit_video', cv2.WINDOW_AUTOSIZE)
```
davidzhao, line 75: similar as above, but probably doesn't matter

livekit-api/livekit/api/version.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.0.1"
```
davidzhao, line 1: give it a bump since it's pretty solid?

```suggestion
__version__ = "0.1.0"
```

davidzhao (approved):
awesome work! just a couple of nits

examples/basic_room.py
```diff
@@ -3,98 +3,98 @@
 from signal import SIGINT, SIGTERM
 from typing import Union
 
-import livekit
+import livekit.rtc as rtc
```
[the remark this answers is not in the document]
theomonnom: lol true

PR #72, opened 2023-10-02
fix: correctly unsubscribe to the ffi queue

theomonnom:
also disable e2ee on publish_wave example

### DIFF
--- a/examples/publish_wave.py
+++ b/examples/publish_wave.py
@@ -1,4 +1,5 @@
 import asyncio
+import time
 import logging
 from signal import SIGINT, SIGTERM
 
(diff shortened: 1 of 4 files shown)


examples/publish_wave.py
```diff
@@ -45,11 +38,9 @@ def on_participant_disconnect(participant: livekit.Participant, *_):
     logging.info("connecting to %s", URL)
     try:
         e2ee_options = livekit.E2EEOptions()
-        e2ee_options.key_provider_options.shared_key = b"password"
 
         await room.connect(URL, TOKEN, options=livekit.RoomOptions(
             auto_subscribe=True,
-            e2ee=e2ee_options
```
davidzhao, line 52: can we leave these two lines there, but just to comment out `e2ee` here.

davidzhao (approved):

examples/publish_wave.py
theomonnom: I think we already have the e2ee.py example for that, I don't think this is necessary

PR #76, opened 2023-10-18
Introduce Agents

keepingitneil:
Agents are loosely defined as server-side participants that consume, process, and/or produce Track data.

This PR introduces the Agent class which is a convenience layer on top of LocalParticipants and Rooms that handles boilerplate for track subscription, participant updates, and more.

Conceptually it is similar to the components-js layer on top of LiveKit's base Javascript client SDK.

### DIFF
--- a/livekit-agents/livekit/__init__.py
+++ b/livekit-agents/livekit/__init__.py
@@ -1,0 +1,1 @@
+from .agent import Agent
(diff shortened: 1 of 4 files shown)


keepingitneil:
Closing, offline discussion led to putting this in its own repo

PR #77, opened 2023-10-23
fix: api improvements

theomonnom:
- remove `pyee` dependency
- renamed `close` to `aclose` for async functions
- added more typing around event listeners
- don't expose any C types 
   - buffer are now copied into the python GC instead 
- setup ruff
   - format & lint
- include typing files into the package


livekit-rtc/livekit/rtc/video_frame.py
```diff
@@ -90,156 +99,346 @@ def create(owned_info: proto_video_frame.OwnedVideoFrameBuffer) \
...
+class ArgbFrame:
     def __init__(self,
+                 data: Union[bytes, bytearray, memoryview],
                  format: VideoFormatType.ValueType,
                  width: int,
-                 height: int) -> None:
+                 height: int,
+                 stride: int = 0) -> None:
+
+        if stride == 0:
```
theomonnom, line 431: Better way to default the stride? 0 looks confusing

livekit-rtc/livekit/rtc/room.py
```diff
@@ -55,7 +59,7 @@ def __init__(self, message: str):
         self.message = message
 
 
-class Room(EventEmitter):
+class Room(EventEmitter[EventTypes]):
```
keepingitneil, line 62: So in this PR we still won't get typing for the Callback args right?

keepingitneil (approved):

livekit-rtc/livekit/rtc/room.py
theomonnom: No, I didn't find a good solution to do it, so atm there is only auto_complete for the type of event

livekit-rtc/livekit/rtc/video_frame.py
```diff
@@ -29,30 +31,38 @@ def __init__(self, timestamp_us: int,
...
     def height(self) -> int:
-        return self._info.height
+        return self._height
 
     @property
     def type(self) -> VideoFrameBufferType.ValueType:
-        return self._info.buffer_type
+        return self._buffer_type
 
+    # TODO(theomonnom): Need Rust modification
```
davidzhao, line 58: still need rust modification?

livekit-rtc/livekit/rtc/video_frame.py
davidzhao: either 0 or None.. IMO a bit of comment would clear it up (i.e. when stride is not passed in, it defaults to size of a single row)

livekit-rtc/livekit/rtc/video_frame.py
```diff
@@ -90,156 +99,346 @@ def create(owned_info: proto_video_frame.OwnedVideoFrameBuffer) \
...
+
+        self._data = bytearray(data)
         self._format = format
-        self.width = width
-        self.height = height
-        self.data = (ctypes.c_uint8 * (width * height *
-                     ctypes.sizeof(ctypes.c_uint32)))()  # alloc frame
+        self._width = width
+        self._height = height
+        self._stride = stride
```
davidzhao, line 441: do you use _stride anywhere?

livekit-api/livekit/api/__init__.py
```diff
@@ -16,6 +16,8 @@
 """
 
 # flake8: noqa
+from ._proto import livekit_room_pb2 as proto_room
```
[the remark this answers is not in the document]
theomonnom: exporting proto_room because there is a lot of nested types needed to build a complete request for e.g

livekit-rtc/livekit/rtc/video_frame.py
theomonnom: working on it

livekit-rtc/livekit/rtc/video_frame.py
theomonnom: on the rust side

livekit-api/livekit/api/__init__.py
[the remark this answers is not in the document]
davidzhao: so users would do import proto_room from livekit-api? that seems strange right?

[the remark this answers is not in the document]
theomonnom: Right, this is not ideal, I'll export every message individually, just afraid of forgetting some, maybe I can just import them all?

davidzhao (approved):
lgtm!

livekit-rtc/livekit/rtc/video_frame.py
```diff
@@ -29,229 +32,575 @@ def __init__(self, timestamp_us: int,
         self.rotation = rotation
 
 
-class VideoFrameBuffer:
-    def __init__(self, owned_info: proto_video_frame.OwnedVideoFrameBuffer) -> None:
-        self._info = owned_info.info
-        self._ffi_handle = FfiHandle(owned_info.handle.id)
+class VideoFrameBuffer(ABC):
```
keepingitneil, line 35: might be worth unit testing all of the different frame types in this one. Seems like there's a fair amount of math that would be hard to debug if it was wrong.

keepingitneil (approved):

PR #98, opened 2023-11-14
add other api services, token verifier & webhooks

theomonnom:
```python
api = api.LivekitAPI("http://theo.livekit.cloud")
api.room.CreateRoom(...)
api.ingress.CreateIngress(...)
```


livekit-api/livekit/api/__init__.py
```diff
@@ -22,6 +22,6 @@
 from livekit.protocol import models
 from livekit.protocol import room
```
davidzhao, line 23: we should re-export the other types too. and maybe change them to this namespace? i.e. so users can call `api.StartWebEgressRequest()` ? versus having to know if a symbol is from room or egress or models

davidzhao (commented):
nice! this completes things quite nicely.

davidzhao (approved):
lgtm!

Perhaps in a separate PR, could you also update README as well?

PR #111, opened 2023-12-04
feat: support topic on data packets

theomonnom:


livekit-rtc/livekit/rtc/participant.py
```diff
@@ -80,7 +80,8 @@ async def publish_data(
         self,
         payload: Union[bytes, str],
         kind: DataPacketKind.ValueType = DataPacketKind.KIND_RELIABLE,
-        destination_sids: Optional[List[Union[str, "RemoteParticipant"]]] = None,
+        destination_sids: List[Union[str, "RemoteParticipant"]] = [],
```
davidzhao, line 83: should we switch to identities now?

theomonnom: Yep, planned

livekit-rtc/livekit/rtc/room.py
```diff
@@ -370,7 +370,13 @@ def _on_room_event(self, event: proto_room.RoomEvent):
             rparticipant = None
             if event.data_received.participant_sid:
                 rparticipant = self.participants[event.data_received.participant_sid]
-            self.emit("data_received", data, event.data_received.kind, rparticipant)
+            self.emit(
+                "data_received",
+                data,
+                event.data_received.kind,
+                rparticipant,
+                event.data_received.topic,
```
davidzhao, line 378: This is problematic to add arguments to callbacks.. my code was just broken with this error:

```
TypeError: on_data_received() takes 4 positional arguments but 5 were given
```

we really need to either:
* avoid changing callback parameters
* pass back an object that can be modified

PR #129, opened 2023-12-21
Replace deprecated pkg_resources

keepingitneil:


theomonnom (approved):
Thanks!

livekit-rtc/livekit/rtc/_ffi_client.py
```diff
@@ -46,13 +45,12 @@ def get_ffi_lib_path():
                 Set LIVEKIT_LIB_PATH to specify a the lib path"
         )
 
-    libpath = pkg_resources.resource_filename(
-        "livekit.rtc", os.path.join("resources", libname)
-    )
-    return libpath
+    with importlib.resources.path("livekit.rtc.resources", libname) as resource_path:
```
theomonnom, line 48: `resources.path` is also deprecated, we need to use `resources.files`

PR #135, opened 2024-01-11
Iterating on Github Action

keepingitneil:


.github/workflows/build-docs.yml
```diff
@@ -34,36 +42,17 @@ jobs:
...
-            echo "No requiress.txt file found."
-          fi
-
-    - name: Debug
-      run: echo $REQUIREMENTS_FILE
-
-    - name: Install Requirements From Egg Info
-      run: python -m pip install -r $REQUIREMENTS_FILE
+      with:
+        submodules: recursive
```
theomonnom, line 46: Do you need submodules?

theomonnom:
Can you try again with the main branch? I changed the build logic

.github/workflows/build-docs.yml
keepingitneil: Yeah - pip install fails without it

keepingitneil:
> * Can you try again with the main branch? I changed the build logic
> 
> Where can I access the docs? Can you add a link to the README

Docs look like this: https://dhoiehpn9ahig.cloudfront.net/livekit-rtc/livekit/rtc.html (still need to figure out cloudfront to so docs.livekit.io redirects to these like the other ones)

.github/workflows/build-docs.yml
```diff
@@ -34,36 +46,20 @@ jobs:
...
 
     - name: Install Package to Document
-      run: python -m pip install -e ${{ inputs.package_dir }}
+      run: python -m pip install ${{ inputs.package_dir }}/
+
+    - name: Download ffi
+      run: "[[ ${{ inputs.package_name }} = 'livekit.rtc' ]] && pip install requests && python livekit-rtc/rust-sdks/download_ffi.py --output $(python -m site --user-site)/livekit/rtc/resources"
```
theomonnom, line 56: Only necessary for livekit-rtc

theomonnom: I think you can simplify this and move to build-rtc.yml

Issue #136, opened 2024-01-12
New RTC Release Errors on Import


Liamazing:
Looks like 0.7.0 errors on import. Here is the output that I get:
```
root@work:/# python3.9 -c "from livekit import rtc"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.9/dist-packages/livekit/rtc/__init__.py", line 31, in <module>
    from .audio_frame import AudioFrame
  File "/usr/local/lib/python3.9/dist-packages/livekit/rtc/audio_frame.py", line 16, in <module>
    from ._ffi_client import FfiHandle, FfiClient
  File "/usr/local/lib/python3.9/dist-packages/livekit/rtc/_ffi_client.py", line 53, in <module>
    ffi_lib = get_ffi_lib()
  File "/usr/local/lib/python3.9/dist-packages/livekit/rtc/_ffi_client.py", line 48, in get_ffi_lib
    with importlib.resources.path("livekit.rtc.resources", libname) as resource_path:
  File "/usr/lib/python3.9/importlib/resources.py", line 165, in path
    _common.as_file(files(package).joinpath(_normalize_path(resource)))
  File "/usr/lib/python3.9/importlib/resources.py", line 147, in files
    return _common.from_package(_get_package(package))
  File "/usr/lib/python3.9/importlib/_common.py", line 14, in from_package
    return fallback_resources(package.__spec__)
  File "/usr/lib/python3.9/importlib/_common.py", line 18, in fallback_resources
    package_directory = pathlib.Path(spec.origin).parent
  File "/usr/lib/python3.9/pathlib.py", line 1072, in __new__
    self = cls._from_parts(args, init=False)
  File "/usr/lib/python3.9/pathlib.py", line 697, in _from_parts
    drv, root, parts = self._parse_args(args)
  File "/usr/lib/python3.9/pathlib.py", line 681, in _parse_args
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```
This output came from a container that was built with the following Dockerfile:
```
FROM ubuntu:20.04
ENV DEBIAN_FRONTEND=noninteractive
RUN apt-get update && apt-get install -y curl python3.9-dev=3.9.5-3* python3-pip
RUN curl -sSL https://get.livekit.io | bash
RUN python3.9 -m pip install --upgrade pip
[...]

PR #138, opened 2024-01-12
don't run builds outside of cibuildwheel

theomonnom:
Fix #134 #129 & don't run the download process inside build_py (only doing it inside cibuildwheel)
Also see https://github.com/livekit/rust-sdks/pull/283


theomonnom:
This should fix https://github.com/livekit/python-sdks/issues/136

livekit-rtc/setup.py
```diff
@@ -33,30 +31,6 @@ def finalize_options(self):
...
-            "python3",
-            str(download_script.absolute()),
-            "--output",
-            str(output.absolute()),
-        ]
-
-        # cibuildwheel is crosscompiling to arm64 on macos, make sure we download the
-        # right binary (kind of a hack here...)
-        if os.environ.get("CIBUILDWHEEL") == "1" and "arm64" in os.environ.get(
-            "ARCHFLAGS", ""
```
theomonnom, line 52: Code moved inside the rust-sdks for simplicity

PR #140, opened 2024-01-12
v0.7.1 & reenable windows

theomonnom:


theomonnom:
disable arm64, it doesn't seem like a priority for now

PR #142, opened 2024-01-13
Relax version pinning for protobuf

keepingitneil:


davidzhao (approved):

livekit-api/livekit/api/version.py
```diff
@@ -1 +1 @@
-__version__ = "0.4.0"
+__version__ = "0.5.0"
```
theomonnom, line 1: Patch instead ?

livekit-protocol/livekit/protocol/version.py
```diff
@@ -1 +1 @@
-__version__ = "0.3.0"
```
theomonnom, line 1: same

PR #143, opened 2024-01-14
ChatManager to manage sending/receiving of chat messages

davidzhao:


davidzhao:
@theomonnom I can't seem to load the shared libs in the test [here](https://github.com/livekit/python-sdks/actions/runs/7521628492/job/20472682809?pr=143#step:4:97). Any ideas?

theomonnom:
We need to download the FFI binaries inside the test workflow. Will do it tmrw

davidzhao:
> We need to download the FFI binaries inside the test workflow. Will do it tmrw
> 
> It is as easy as executing the download-ffi.py script to the resources path

that worked. Thanks!

livekit-rtc/livekit/rtc/chat.py
```diff
@@ -0,0 +1,140 @@
...
+
+from .room import Room, Participant, DataPacket
+from ._proto.room_pb2 import DataPacketKind
+from ._utils import generate_random_base62
+
+_CHAT_TOPIC = "lk-chat-topic"
+_CHAT_UPDATE_TOPIC = "lk-chat-update-topic"
+
+
+class ChatManager:
```
keepingitneil, line 29: Should this class also accumulate chat messages with a sane default way of updating chat messages (i.e. setting deleted from false to true)?

keepingitneil (approved):

keepingitneil:
@davidzhao @theomonnom I did it here as well. Maybe this can be a shared workflow step: https://github.com/livekit/python-sdks/pull/135/files#diff-4fdc48ab490b670c90974725b8e3d399b24c03d6d09563acf8ea87d4113930fbR55

livekit-rtc/livekit/rtc/chat.py
davidzhao: yeah I def want to in the future.. just wanted to keep it simple because if we started buffering, we'd need a way to clean it up too to avoid it buffering forever

theomonnom:
> @davidzhao @theomonnom I did it here as well. Maybe this can be a shared workflow step: https://github.com/livekit/python-sdks/pull/135/files#diff-4fdc48ab490b670c90974725b8e3d399b24c03d6d09563acf8ea87d4113930fbR55

No need, it's one line. I think you're complicating it inside your build_docs workflow. Only livekit-rtc needs it

livekit-rtc/livekit/rtc/chat.py
```diff
@@ -0,0 +1,140 @@
...
+            timestamp = datetime.fromtimestamp(d.get("timestamp") / 1000.0)
+        msg = cls(
+            id=id,
+            timestamp=timestamp,
+        )
+        msg.update_from_jsondict(d)
+        return msg
+
+    def update_from_jsondict(self, d: Dict[str, Any]) -> None:
+        self.message = d.get("message")
```
theomonnom, line 128: if message is None, should we default to the current message?
d.get("message", self.message)

livekit-rtc/livekit/rtc/chat.py
```diff
@@ -0,0 +1,140 @@
...
+        If message.deleted is set to True, we'll signal to remote participants that the message
+        should be deleted.
+        """
+        await self._lp.publish_data(
+            payload=json.dumps(message.asjsondict()),
+            kind=DataPacketKind.KIND_RELIABLE,
+            topic=_CHAT_UPDATE_TOPIC,
+        )
+
+    def on_message(self, callback: Callable[["ChatMessage"], None]):
```
theomonnom, line 82: We don't have on_message_updated?
IMO let's use https://github.com/livekit/python-sdks/blob/main/livekit-rtc/livekit/rtc/_event_emitter.py since we're already using it on other places

livekit-rtc/livekit/rtc/chat.py
davidzhao: no, the protocol does not support delta updates. it's valid for message to be updated to None.. in the case of deletion.

livekit-rtc/livekit/rtc/chat.py
davidzhao: will do

theomonnom:
Screenshot 2024-01-15 at 9 24 50 AM
We don't have CI yet, but here are some types complains with mypy

davidzhao:
@theomonnom I don't have mypy setup.. It'll be faster if you can please fix lint issues directly. I'd like to get this released so we can update everything else

theomonnom (approved):
done, lgtm!

theomonnom:
nvm last thing is: do we need on_message_updated?
Ok to merge before anyway

davidzhao:
> nvm last thing is: do we need on_message_updated? Ok to merge before anyway

Thanks! we don't need it right now. Will add this once finalized across clients

PR #144, opened 2024-01-15
Improvements to VideoFrame and publish_track APIs

davidzhao:
Using defaults instead of requiring users to pass in these parameters.

This *is* a breaking change. We'll need to release 0.8.0 afterwards.

### DIFF
--- a/livekit-rtc/livekit/rtc/video_frame.py
+++ b/livekit-rtc/livekit/rtc/video_frame.py
@@ -26,10 +26,18 @@ from abc import ABC, abstractmethod
 class VideoFrame:
     def __init__(
         self,
-        timestamp_us: int,
-        rotation: VideoRotation.ValueType,
         buffer: "VideoFrameBuffer",
+        *,
+        timestamp_us: int = 0,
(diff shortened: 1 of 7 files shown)


livekit-rtc/livekit/rtc/video_frame.py
```diff
@@ -26,10 +26,18 @@
 class VideoFrame:
     def __init__(
         self,
-        timestamp_us: int,
-        rotation: VideoRotation.ValueType,
         buffer: "VideoFrameBuffer",
+        *,
```
theomonnom, line 30: This is a breaking change

theomonnom (approved):
We could avoid the breaking change by removing the *. Tho still nit
otherwise lgtm

davidzhao:
> We could avoid the breaking change by removing the *. Tho nit otherwise lgtm, ty for the docstr

the order of args is still different though.. so I think it'd break either way? Just thinking that it'll only get harder to modify the API.. so might as well do it now 😛

theomonnom:
Oh right, didn't see it

PR #166, opened 2024-02-14
remove whisper example

theomonnom:
Not maintained and whisper.cpp changed, I think it is more worth to add it inside agents in the future


davidzhao (approved):

neryabCommitted:
hey, is there a work being done to add it inside of agents?

PR #174, opened 2024-03-19
Use protoc's --pyi_out instead of mypy-protobuf package

keepingitneil:
And add type checking GHA

Successful type checking run: https://github.com/livekit/python-sdks/actions/runs/8351262508

Purposely failed type checking run by adding `bad_variable: str = 5` to a few files: https://github.com/livekit/python-sdks/actions/runs/8351236763


livekit-api/livekit/api/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2023 LiveKit, Inc.
+# Copyright 2024 LiveKit, Inc.
```
davidzhao, line 1: don't need to change this

davidzhao (approved):

theomonnom:
Is changing mypy to pyi generator a breaking change?

livekit-rtc/livekit/rtc/room.py
```diff
@@ -80,9 +81,9 @@ class RoomOptions:
 class DataPacket:
     data: bytes
     kind: proto_room.DataPacketKind.ValueType
```
theomonnom, line 83: Weird thing is that ValueType doesn't exist anymore due to the pyi generator but mypy is not complaining ? Seems like the type checking isn't working?

theomonnom:
The mypy generator may be a better choice for stricter type checking

livekit-rtc/livekit/rtc/room.py
keepingitneil: Oh this is Room is not from livekit-protocol, it's from livekit/rtc/_proto which is still generated by mypy-protobuf.

keepingitneil: Just to double check, changed kind to something bad (i.e. `str`) and ran it again and it failed as expected: https://github.com/livekit/python-sdks/actions/runs/8362789419

keepingitneil:
> The mypy generator may be a better choice for stricter type checking if the types are not detected (the above comment)

Yeah in my testing pyi_out has been working flawlessly. Since it's newer, a bit more "official", and  removes a build dependency I decided to make the switch. I left livekit-rtc in-tact but we can switch to pyi_out for those protos too in another PR if you're satisfied with the pyi_out typing.

keepingitneil:
> Is changing mypy to pyi generator a breaking change?

It should not be. It only affects the pyi files. The generated py files remain as-is.

theomonnom (approved):

theomonnom (approved):
nice!

PR #190, opened 2024-04-09
Add SIP services. Support SIP DTMF in RTC.

dennwc:
Requires https://github.com/livekit/rust-sdks/pull/319


livekit-protocol/livekit/protocol/agent.py
```diff
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
 # source: livekit_agent.proto
-# Protobuf Python Version: 4.25.3
```
dennwc: I noticed that my protoc generates this diff. Is my version older or newer than the one used in the SDK?

PR #194, opened 2024-04-26
Introduce transcription protocol

keepingitneil:


davidzhao (approved):
lgtm!

thanks for fixing the CI too

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -117,6 +131,42 @@ async def publish_data(
         if cb.publish_data.error:
             raise PublishDataError(cb.publish_data.error)
 
+    async def publish_transcription(
+        self,
+        participant_identity: str,
```
theomonnom, line 136: I would be more comfortable creating a Transcription structure here.
On the agents API, we will need to return a Transcription, and I don't think this should be defined inside agents.

publish_transcription accepting a "Transcription" seems OK to me

livekit-rtc/livekit/rtc/transcription.py
```diff
@@ -0,0 +1,10 @@
+from dataclasses import dataclass
+
+
```
theomonnom, line 3: Add it here?

theomonnom (changes requested):
I'm also wondering if start_time and end_time should be float on the Python side.
To respect the time.time() format.

Maybe we could automatically do the conversion?

keepingitneil:
@theomonnom re: float. I think we should keep it int because the clock being used will be the frame clock. Although I'm wondering if we should specify units in the transcription protocol now: `start_time_us` ?

theomonnom (approved):

theomonnom:
let's mb wait for a release tho?

PR #211, opened 2024-06-20
Update protocol. Support SIP inbound/outbound trunks, grants.

dennwc:
Update protocol. Add support for inbound/outbound SIP trunks and set token grants.

Please check the token construction, I'm not sure I added everything required for new `sip` claim field.


theomonnom (approved):
nice!

livekit-api/livekit/api/_service.py
```diff
@@ -15,8 +15,14 @@ def __init__(
         self.api_key = api_key
         self.api_secret = api_secret
 
-    def _auth_header(self, grants: VideoGrants) -> Dict[str, str]:
-        token = AccessToken(self.api_key, self.api_secret).with_grants(grants).to_jwt()
+    def _auth_header(
+        self, grants: VideoGrants, sip: SIPGrants = None
```
theomonnom, line 19:
```suggestion
        self, grants: VideoGrants, sip: SIPGrants | None = None
```

theomonnom:
Type checks aren't passing

livekit-api/livekit/api/_service.py
dennwc: After adding it, the type checker complains about requiring Python 3.10. So it's probably worth doing this update separately.

dennwc:
It's weird that the checks do not appear on the PR :thinking:

livekit-api/livekit/api/_service.py
dennwc: Uh, no, type check doesn't like it this way either... Let's see how to bump to Python 3.10.

dennwc:
Should be good now. @theomonnom could you double check?

livekit-api/livekit/api/_service.py
```diff
@@ -2,7 +2,7 @@
 import aiohttp
```
theomonnom, line 2: Can you add  from `__future__ import __annotations__` in the beginning of the file?
This allows to use Union syntax like `SIPGrants | None` on older python versions

dennwc: Should I rollback Python version?

dennwc: @theomonnom I tried to rollback the version, but it doesn't recognize the future import. So I'll keep Python 3.10 then.

davidzhao: I think it should be `from __future__ import annotations` ?

dennwc: That works, thank you!

PR #215, opened 2024-07-06
room: allow RemoteParticipant to be None

nbsp:
should fix the bug when receiving messages from the server: this allows Python to fail gracefully to retrieve the remote participant.

### DIFF
--- a/livekit-rtc/livekit/rtc/room.py
+++ b/livekit-rtc/livekit/rtc/room.py
@@ -405,9 +405,12 @@ class Room(EventEmitter[EventTypes]):
 
                 data = bytes(native_data)
                 FfiHandle(owned_buffer_info.handle.id)
-                rparticipant = cast(RemoteParticipant, self._retrieve_remote_participant(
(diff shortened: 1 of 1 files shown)


davidzhao (approved):

davidzhao:
looks like we need to update the return type too

theomonnom (approved):
Lgtm, CI doesn't pass tho

nbsp:
> looks like we need to update the return type too

where? the return type on this specific function is Optional[] now, and as far as i can tell the rest are already like that too

nbsp:
unsure why this is still erroring, i've got `assert isinstance` guards there..

davidzhao:
> unsure why this is still erroring, i've got `assert isinstance` guards there..

have you tried `isinstance(obj, Participant)` or `if obj is not None` ?

nbsp:
> > unsure why this is still erroring, i've got `assert isinstance` guards there..
> 
> have you tried `isinstance(obj, Participant)` or `if obj is not None` ?

found a fix by using typing.cast

PR #217, opened 2024-07-12
Update rust ffi

keepingitneil:


livekit-rtc/livekit/rtc/room.py
```diff
@@ -112,16 +113,19 @@ def __init__(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> None:
         self._info = proto_room.RoomInfo()
 
         self.participants: Dict[str, RemoteParticipant] = {}
```
theomonnom, line 115: new api is using remote_participants I think

theomonnom: let's make sure we're aligned with other SDKs, we can only break once if we release v1

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -60,7 +60,7 @@ class Participant:
     def __init__(self, owned_info: proto_participant.OwnedParticipant) -> None:
         self._info = owned_info.info
         self._ffi_handle = FfiHandle(owned_info.handle.id)
-        self.tracks: dict[str, TrackPublication] = {}
+        self.trackPublications: dict[str, TrackPublication] = {}
```
davidzhao, line 63:
```suggestion
        self.track_publications: dict[str, TrackPublication] = {}
```

keepingitneil: realized this about 0.2 seconds after I pushed. fixed

livekit-rtc/livekit/rtc/room.py
```diff
@@ -265,35 +269,40 @@ def _on_room_event(self, event: proto_room.RoomEvent):
...
+                event.track_published.participant_identity
+            ]
             rpublication = RemoteTrackPublication(event.track_published.publication)
-            rparticipant.tracks[rpublication.sid] = rpublication
+            rparticipant.track_publications[rpublication.sid] = rpublication
             self.emit("track_published", rpublication, rparticipant)
         elif which == "track_unpublished":
-            rparticipant = self.participants[event.track_unpublished.participant_sid]
-            rpublication = rparticipant.tracks.pop(
+            rparticipant = self.remote_participants[
```
davidzhao, line 292: nit: I think we should always use .get() on these to guard against the object not being present. if participant is removed, it'd be better to handle the None responses.

davidzhao (approved):

livekit-rtc/livekit/rtc/room.py
keepingitneil: @theomonnom is it possible for track_unpublished (and friends) to come after the participant has been disconnected. I'm not super familiar with the async rust stuff and what guarantees we have for ffi events.

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -88,13 +91,14 @@ def __init__(
     ) -> None:
         super().__init__(owned_info)
         self._room_queue = room_queue
-        self.tracks: dict[str, LocalTrackPublication] = {}  # type: ignore
+        self.track_publications: dict[str, LocalTrackPublication] = {}  # type: ignore
 
     async def publish_data(
         self,
         payload: Union[bytes, str],
```
theomonnom, line 98: should payload be positional?

theomonnom (approved):
awesome lgtm! 🔥

livekit-rtc/livekit/rtc/participant.py
keepingitneil: I like it positional because it's the only value with no default so it's really convenient

PR #229, opened 2024-08-01
Introduce AudioStream.from_participant + VideoStream.from_participant

keepingitneil:
This PR introduces the following APIs:
`AudioStream.from_participant` and `VideoStream.from_participant`


livekit-rtc/livekit/rtc/video_stream.py
```diff
@@ -37,34 +41,81 @@ class VideoStream:
     def __init__(
         self,
         track: Track,
-        *,
         loop: Optional[asyncio.AbstractEventLoop] = None,
         capacity: int = 0,
         format: Optional[proto_video_frame.VideoBufferType.ValueType] = None,
+        **kwargs,
```
theomonnom, line 47: I think it's OK to directly put participant here too?
VideoStream.from_participant and AudioStream.from_participant are not necessary I think. We can use typing.overload on the constructor?

keepingitneil: I prefer from_participant because it's the only one that uses track_source as well. So track_source and participant are used for the participant path and track is used for the other path. This is also why I chose to use kwargs (so it doesn't show up as an option during autocomplete) although I'll ready more about typing.overload to see if that could be useful here

keepingitneil: overloading the types is kind of clunky here because the args are positional and there are default values. So in the second overload I would need to enforce kwargs with `*` while keeping the original overload in tact. Do you still think it's better to remove `from_participant` in favor of `__init__` overloads? My personal preference is still the from_participant

keepingitneil: offline discussion. Keep as-is + introduce `from_track` for symmetry

livekit-rtc/livekit/rtc/audio_stream.py
```diff
@@ -39,30 +43,93 @@ def __init__(
         capacity: int = 0,
         sample_rate: int = 48000,
         num_channels: int = 1,
+        **kwargs,
     ) -> None:
         self._track = track
```
theomonnom, line 48:
```suggestion
        self._track: Track | None = track
```

theomonnom (approved):

PR #256, opened 2024-09-19
only subtract 50ms from stream queue once

nbsp:
resetting to -0.05 on release also breaks things, so it was left at 0.0.

### DIFF
--- a/livekit-rtc/livekit/rtc/audio_source.py
+++ b/livekit-rtc/livekit/rtc/audio_source.py
@@ -57,7 +57,10 @@ class AudioSource:
         self._info = resp.new_audio_source.source
         self._ffi_handle = FfiHandle(self._info.handle.id)
 
-        self._last_capture = self._q_size = 0.0
+        # remove 50ms to account for processing time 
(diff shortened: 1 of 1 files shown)


livekit-rtc/livekit/rtc/audio_source.py
```diff
@@ -57,7 +57,10 @@ def __init__(
         self._info = resp.new_audio_source.source
         self._ffi_handle = FfiHandle(self._info.handle.id)
 
-        self._last_capture = self._q_size = 0.0
+        # remove 50ms to account for processing time 
+        # (e.g. using wait_for_playout for very small chunks)
+        self._q_size = -0.05
```
theomonnom, line 62: This isn't going to work because it assumes we're always sending data.

What is actually being done is:
```
capture(10seconds)
waitForPlayout(10seconds-50ms)

sleep(20seconds)

capture(10seconds)
waitForPlayout(10seconds-50ms
```)

nbsp: is this approach better? subtracting directly from the timeout without removing it from the total tally

PR #263, opened 2024-09-26
add with_kind to claims

### LINKED ISSUE #262: AccessToken is missing with_kind
Can't generate an agent access token which is a problem for certain operations like transcript publication which require agent permissions

timu-jesse-ezell:
Fixes #262


davidzhao (approved):
looks good! thanks!

PR #273, opened 2024-10-02
rtc: added sip publish dtmf feature

s-hamdananwar:


livekit-rtc/livekit/rtc/participant.py
```diff
@@ -147,6 +152,34 @@ async def publish_data(
         if cb.publish_data.error:
             raise PublishDataError(cb.publish_data.error)
 
+    async def publish_dtmf(self, code: int, digit: str) -> None:
```
theomonnom, line 155:
```suggestion
    async def publish_dtmf(self, *, code: int, digit: str) -> None:
```
Let's add a "*" to be make the code using this fnc more explicit

theomonnom (approved):
Nice lgtm!

PR #283, opened 2024-10-08
[WIP] RPC

bcherry:


lukasIO (approved):

examples/rpc.py
```diff
@@ -0,0 +1,227 @@
...
+        result = dividend / divisor
+        return json.dumps({"result": result})
+
+    await greeters_room.local_participant.register_rpc_method("arrival", arrival_method)
+    await math_genius_room.local_participant.register_rpc_method(
+        "square-root", square_root_method
+    )
+    await math_genius_room.local_participant.register_rpc_method(
+        "divide", divide_method
+    )
```
theomonnom, line 113: does registering needs to be async?

bcherry: yeah this PR needs some small updates on latest FFI once the protobuf issues are resolved so i can test it. i'll clean this up when i do

bcherry: done

bcherry:
@theomonnom this has been updated to remove the unneeded asyncs (based on FFI refinement you helped with earlier this week) and the addition of docs in the README.

examples/rpc.py
```diff
@@ -0,0 +1,219 @@
...
+    print("Participants disconnected. Example completed.")
+
+
+def register_receiver_methods(greeters_room: rtc.Room, math_genius_room: rtc.Room):
+    @greeters_room.local_participant.rpc_method("arrival")
+    async def arrival_method(
+        request_id: str,
+        caller_identity: str,
+        payload: str,
+        response_timeout_ms: int,
```
theomonnom, line 64: I think we should put that inside a dataclass like RpcData.
Otherwise it isn't future proof and may break if we want to add a new field in the future

theomonnom: Ppl may also ignore some args, and Python will raise an exception when trying to call a function with more arguments than what the function defines

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -236,6 +241,183 @@ async def publish_transcription(self, transcription: Transcription) -> None:
...
+            response_error = RpcError._built_in(RpcError.ErrorCode.UNSUPPORTED_METHOD)
+        else:
+            try:
+                response_payload = await handler(
+                    request_id, caller_identity, payload, response_timeout_ms
+                )
+            except RpcError as error:
+                response_error = error
+            except Exception as error:
+                print(
```
theomonnom, line 397: We shouldn't use print, use logger.exception

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -236,6 +241,183 @@ async def publish_transcription(self, transcription: Transcription) -> None:
...
+                invocation_id=invocation_id,
+                error=response_error.to_proto() if response_error else None,
+                payload=response_payload,
+            )
+        )
+
+        res = FfiClient.instance.request(req)
+
+        if res.rpc_method_invocation_response.error:
+            print(
```
theomonnom, line 417: raise instead?

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -236,6 +241,183 @@ async def publish_transcription(self, transcription: Transcription) -> None:
...
+        response_error: Optional[RpcError] = None
+        response_payload: Optional[str] = None
+
+        handler = self._rpc_handlers.get(method)
+
+        if not handler:
+            response_error = RpcError._built_in(RpcError.ErrorCode.UNSUPPORTED_METHOD)
+        else:
+            try:
+                response_payload = await handler(
```
theomonnom, line 391: Should we allow non-async handlers too?

livekit-rtc/livekit/rtc/room.py
```diff
@@ -436,6 +439,28 @@ async def _listen_task(self) -> None:
             self._room_queue.put_nowait(event)
             await self._room_queue.join()
 
+    def _on_rpc_method_invocation(self, rpc_invocation: RpcMethodInvocationEvent):
+        if self._local_participant is None:
+            logging.warning(
+                "Received RPC invocation before local participant was initialized"
+            )
+            return
```
theomonnom, line 447: you can safely remove this code

livekit-rtc/livekit/rtc/room.py
```diff
@@ -436,6 +439,28 @@ async def _listen_task(self) -> None:
...
+
+        if (
+            rpc_invocation.local_participant_handle
+            == self._local_participant._ffi_handle.handle
+        ):
+            asyncio.create_task(
```
theomonnom, line 453: We need to keep a reference to the tasks. (See https://docs.python.org/3.12/library/asyncio-task.html#asyncio.create_task)
It's also unclear when we should cancel/terminate them (at room shutdown?)

Can you also explicitly use the loop of the room (self._loop)

theomonnom (changes requested):

livekit-rtc/livekit/rtc/participant.py
bcherry: yeah I think its reasonable. I changed the handler return type to `Union[Awaitable[str], str]` and then in this spot i checked it with `if asyncio.iscoroutinefunction(handler):` before awaiting.

livekit-rtc/livekit/rtc/room.py
bcherry: Ok so I did this and went ahead and cancelled them on room disconnect. I also added tests for this behavior and timeouts in the example and found a couple small issues that I addressed. I also decided to switch response_timeout_ms to response_timeout in float seconds to be more pythonic and I'll make the a similar change in node and JS to drop the `ms` part of the name and be more platform-specific.

examples/rpc.py
bcherry: I'm not really sure a dataclass is better. it adds more indirection to the documentation and increases the surface area of imports needed. 

Why not just make these kwargs only? I changed it so you must call this method with kwargs (by adding a `*` before them, which works in python 3). If we introduce new ones in the future that's fine, we can ensure they have a default value which is the same as with a data class.

What do you think?

livekit-rtc/livekit/rtc/room.py
bcherry: actually i couldn't remove this because the type checker complains about accessing properties if its none. i removed the log line but kept the check

examples/rpc.py
theomonnom: What I mean is that if if the user defines its callback this way:

```
def my_callback(arg1, arg2):
   pass

```

and in our code we call the callback with 3 args `my_callback(arg1, arg2, arg3)`. Python will raise an exception. So the code isn't future-proof since as soon as we add a new argument, it'll break

bcherry: oh you mean on the handler - ok will take a look

bcherry: this was a good call - made that change and will go make it everywhere else as well

livekit-rtc/livekit/rtc/room.py
```diff
@@ -399,9 +401,13 @@ async def disconnect(self) -> None:
         if not self.isconnected():
             return
 
+        if self._rpc_invocation_tasks:
+            for task in self._rpc_invocation_tasks:
+                task.cancel()
+            await asyncio.gather(*self._rpc_invocation_tasks, return_exceptions=True)
```
theomonnom, line 407: This is only being executed when the user is explicitly calling the `disconnect` method. You should mode that at the ned of the main task

theomonnom (approved):
lgtm!

nit: CI is failing

livekit-rtc/livekit/rtc/rpc.py
```diff
@@ -0,0 +1,109 @@
...
+        Creates an error object with the given code and message, plus an optional data payload.
+
+        If thrown in an RPC method handler, the error will be sent back to the caller.
+
+        Error codes 1001-1999 are reserved for built-in errors (see RpcError.ErrorCode for their meanings).
+        """
+        super().__init__(message)
+        self.code = code
+        self.message = message
+        self.data = data
```
theomonnom, line 92: We should prefer to expose those using @property. Otherwise they're not referenced on the docs and they're mutable

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -236,6 +245,204 @@ async def publish_transcription(self, transcription: Transcription) -> None:
...
+            `register_rpc_method` for more details
+        """
+
+        def decorator(
+            handler: Callable[[RpcInvocationData], Union[Awaitable[str], str]],
+        ):
+            self.register_rpc_method(method, handler)
+            return handler
+
+        return decorator
```
theomonnom, line 358: I think rpc_method and register_rpc_method should be the same method

theomonnom: Example on how I do it for the EventEmitter: https://github.com/livekit/python-sdks/blob/5b78bcc816d9254be7b82421be2d18b80531335a/livekit-rtc/livekit/rtc/event_emitter.py#L119

livekit-rtc/livekit/rtc/room.py
bcherry: That makes sense. I moved it to the end of `_listen_task` but I found that it no longer has the expected behavior in my example app when disconnect is called manually. So I left it in both places, I think the result will be a little messier and the caller might get the wrong error if the recipient gets forcibly disconnected but it should be ok for now.

livekit-rtc/livekit/rtc/participant.py
bcherry: Yeah I was wondering if that's possible - I'd definitely prefer it that way. Thanks for the example!

livekit-rtc/livekit/rtc/rpc.py
bcherry: Good call, thank you.

PR #291, opened 2024-10-17
Send SDK override and version to FFI

bcherry:
This adds the new required params for SDK and SDK version, as added in https://github.com/livekit/rust-sdks/pull/471

It looks like this works but it's stuck on a different protobuf error during connection so I haven't verified they do make it through unharmed to the SFU but I believe it will work.


livekit-rtc/livekit/rtc/_ffi_client.py
```diff
@@ -54,14 +55,20 @@ def get_ffi_lib():
     res = importlib.resources.files("livekit.rtc.resources") / libname
     ctx = importlib.resources.as_file(res)
     path = _resource_files.enter_context(ctx)
+    print(path)
```
theomonnom, line 58:
```suggestion
```

livekit-rtc/livekit/rtc/_ffi_client.py
```diff
@@ -204,8 +211,13 @@ def instance(cls) -> "FfiClient":
...
         self._lock = threading.RLock()
         self._queue = FfiQueue[proto_ffi.FfiEvent]()
-
-        ffi_lib.livekit_ffi_initialize(ffi_event_callback, True)
+        
+        ffi_lib.livekit_ffi_initialize(
+            ffi_event_callback,
+            True,
+            b"python",
+            version("livekit").encode('utf-8')
```
theomonnom, line 219: let's use our version.py as the ground truth 
from ..version import __version__

theomonnom (approved):
otherwise lgtm

livekit-rtc/livekit/rtc/_ffi_client.py
```diff
@@ -205,7 +211,9 @@ def __init__(self) -> None:
         self._lock = threading.RLock()
         self._queue = FfiQueue[proto_ffi.FfiEvent]()
 
-        ffi_lib.livekit_ffi_initialize(ffi_event_callback, True)
+        ffi_lib.livekit_ffi_initialize(
+            ffi_event_callback, True, b"python", version("livekit").encode("utf-8")
```
theomonnom, line 215: encode should be ascii

livekit-rtc/livekit/rtc/_ffi_client.py
bcherry: ah - thank you! didn't see that

livekit-rtc/livekit/rtc/_ffi_client.py
bcherry: oops, fixed

livekit-rtc/livekit/rtc/_ffi_client.py
bcherry: makes sense - updated

PR #293, opened 2024-10-18
eventemitter should raise TypeError when there isn't enough arguments passed on emit

theomonnom:


davidzhao (commented):
if we are passing 3 args, but user only needs two, it seems like we could make it work by passing in less?

theomonnom:
> if we are passing 3 args, but user only needs two, it seems like we could make it work by passing in less?
it is the case

livekit-rtc/tests/test_emitter.py
```diff
@@ -0,0 +1,103 @@
...
+    emitter = EventEmitter[EventTypes]()
+
+    calls = []
+
+    @emitter.on("whatever")
+    def on_whatever(first, second, third):
+        calls.append((first, second, third))
+
+    emitter.emit("whatever", 1, 2, 3)
+    emitter.emit("whatever", 1, 2, 3, 4, 5)  # only 3 arguments will be passed
```
theomonnom, line 61: check here

davidzhao (approved):

PR #302, opened 2024-11-03
update datetime.utcnow() to use UTC timezone-aware object

lebaudantoine:
Following deprecation warning was raised:
```
  /usr/local/lib/python3.12/site-packages/livekit/api/access_token.py:152: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    (datetime.datetime.utcnow() + self.ttl).utctimetuple()
```

datetime.utcnow() is being deprecated in Python 3.12 as it returns a naive datetime object without timezone information. Replace with datetime.now(UTC) to use an explicit timezone-aware object, preventing potential timezone ambiguity issues.

In access_token.py, updated the token expiration calculation to use the recommended timezone-aware approach.

### DIFF
--- a/livekit-api/livekit/api/access_token.py
+++ b/livekit-api/livekit/api/access_token.py
@@ -154,9 +154,9 @@ class AccessToken:
             {
                 "sub": self.identity,
                 "iss": self.api_key,
-                "nbf": calendar.timegm(datetime.datetime.utcnow().utctimetuple()),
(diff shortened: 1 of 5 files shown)


davidzhao (approved):
lgtm

davidzhao:
thanks for the fix!

PR #303, opened 2024-11-06
Agent dispatch APIs, ability to set room config in token

davidzhao:


livekit-api/livekit/api/access_token.py
```diff
@@ -75,12 +77,28 @@ class SIPGrants:
...
+    metadata: str = ""
+    video: Optional[VideoGrants] = None
+    sip: Optional[SIPGrants] = None
+    attributes: Optional[dict[str, str]] = None
+    sha256: Optional[str] = None
+    room_preset: Optional[str] = None
+    room_config: Optional[RoomConfiguration] = None
+
+    def asdict(self) -> dict:
+        # we don't
```
theomonnom, line 90:
```suggestion
```

livekit-api/livekit/api/agent_dispatch_service.py
```diff
@@ -0,0 +1,108 @@
...
+            SVC,
+            "ListDispatch",
+            proto_agent_dispatch.ListAgentDispatchRequest(room=room_name),
+            self._auth_header(VideoGrants(room_admin=True, room=room_name)),
+            proto_agent_dispatch.ListAgentDispatchResponse,
+        )
+        return list(res.agent_dispatches)
+
+    async def get_dispatch(
+        self, dispatch_id: str, room_name: str
```
theomonnom, line 86: For consistency, wdyt of using `proto_agent_dispatch.ListAgentDispatchRequest` as the parameter. (not a big deal otherwise)

davidzhao: I think here i'm mainly trying to make the API easier to understand. it's always kind of confusing seeing a `List*` API for when you are looking to retrieve a single element.

davidzhao:
updated, PTAL

livekit-api/livekit/api/agent_dispatch_service.py
theomonnom: For other services, we don't expose any get_ method.
Only list (e.g [egress](https://github.com/livekit/python-sdks/blob/e18187079b4706d78ed70fcecd5124f66eaa5e9c/livekit-api/livekit/api/egress_service.py#L92))

Maybe that's fine but adding new parameters in the future may be a pain. So if we go into this path, can you make them kwarg

davidzhao: we didn't have it for the other services, and I think that's a shortcoming. optimizing for user clarity, `get` > `list` when getting a single entry.

I'm not sure that it makes sense to make the required args as kwargs. what is the motivation there? we can always add kwargs for optional args as kwargs.

theomonnom: It makes the code more explicit (e.g which position is the room name, first or second?)

theomonnom: I generally use positional argument when there is only one **main** arguement. Tho no strong opinion

davidzhao: I see what you mean, especially when it gets longer, but 2 seems very straight forward IMHO.

nbsp: fwiw it's still possible to treat positional arguments as named in python. this just allows you to do either/or

theomonnom: I generally max to 1, but no strong opinion, lgtm

PR #306, opened 2024-11-13
Improve reference docs for Server APIs

bcherry:
I found this module to be quite difficult to use within Agents due to limited documentation.

This PR:

- Adds a clearer top-level explanation
- Hides some red-herring modules
- Explicitly points towards using LiveKitAPI
- Adds return types and cross links LiveKitAPI.room, egress, etc  to their documentation
- Adds full docs for the RoomService client requests (I left others alone for now)
- Removes the `proto` imports that were misleading in the docs/source as the right way to import the proto classes is via api.* (they are re-exported in `__init__.py`)
- Add `__aenter__` and `__aexit__` to LiveKitAPI so it can be used with `async with` which allows cleaner example code (lmk @theomonnom if I did this wrong, I'm not super familiar)


davidzhao (approved):
nice!

PR #311, opened 2024-11-20
added explicit agent dispatch example

s-hamdananwar:


examples/agent_dispatch.py
```diff
@@ -0,0 +1,74 @@
...
+room_name = "my-room"
+agent_name = "test-agent"
+
+"""
+This example demonstrates how to have an agent join a room 
+without using the automatic dispatch. In order to use this 
+feature, you must have an agent running with `agent_name` set 
+when defining your WorkerOptions. A dispatch requests the 
+agent to enter a specific room with optional metadata.
+"""
```
bcherry, line 22: link to the docs you wrote too

bcherry (approved):

examples/agent_dispatch.py
s-hamdananwar: the docs is still not published as DZ is taking it over and creating a dedicated docs page for dispatching. will link to it once that is out

PR #312, opened 2024-11-21
Raise error if EventEmitter used with async callback

bcherry:
Some reasonable feedback arrived in community that it's confusing that you can't use async handlers for events in this SDK.  

I looked at the complexity of supporting async handlers and it seems like the largest barrier would be how to await them?  Either the `.emit()` method would need to become async and await them directly, which sounds wrong as you wouldn't expect `.emit()` to block on long-running applications, or you'd need to create tasks to run them asynchronously but then you have the question of where to collect the tasks and how to cancel them or cleanup.

I think it's reasonable to pass this onto the end-developer to collect their own async tasks but we can make it easier to discover this and harder to write subtle bugs. Raising an error with an explanation and workaround suggestion if you pass an async callback seems like a good start.

### DIFF
--- a/livekit-rtc/tests/test_emitter.py
+++ b/livekit-rtc/tests/test_emitter.py
@@ -51,35 +51,30 @@ def test_args():
 
     emitter = EventEmitter[EventTypes]()
 
-    calls = []
+    args_calls = []
 
-    @emitter.on("whatever")
+    @emitter.on("whatever_args")
     def on_whatever(first, second, third):
-        calls.append((first, second, third))
+        args_calls.append((first, second, third))
 
-    emitter.emit("whatever", 1, 2, 3)
-    emitter.emit("whatever", 1, 2, 3, 4, 5)  # only 3 arguments will be passed
+    emitter.emit("whatever_args", 1, 2, 3)
+    emitter.emit("whatever_args", 1, 2, 3, 4, 5)  # only 3 arguments will be passed
 
-    assert len(calls) == 2
-    assert calls[0] == (1, 2, 3)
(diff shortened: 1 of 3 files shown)


nbsp (approved):
is there no way to do this within python's typing instead of throwing our own error? otherwise lg

nbsp:
i think next steps would be to detect whether it's a coro and wrap it ourselves, because i agree the method of doing it right now is kind of cumbersome

bcherry:
> is there no way to do this within python's typing instead of throwing our own error? otherwise lg

AFAIK python doesn't have any runtime type checking facilities. We do provide the necessary type hints but I think it requires some special configuration to get your local IDE to pull them through correctly.

theomonnom (approved):

livekit-rtc/tests/test_emitter.py
```diff
@@ -60,26 +60,27 @@ def on_whatever(first, second, third):
...
     def on_whatever_varargs(*args):
         calls.append(args)
 
     emitter.emit("whatever", 1, 2, 3, 4, 5)
+    emitter.emit("whatever", 1, 2)
 
-    assert len(calls) == 2
-    assert calls[0] == (1, 2, 3)
-    assert calls[1] == (1, 2, 3, 4, 5)
+    assert calls == [(1, 2, 3, 4, 5), (1, 2)]
```
bcherry, line 83: This test was flaky because callbacks are stored in a `set` and thus aren't explicitly ordered, and this check relied on a particular callback order to work.

I broke the test in two (one for regular args and one for varargs) and fixed the flakiness

PR #319, opened 2024-12-06
fixed agent_dispatch.py typos

s-hamdananwar:


kath0la (approved):
Thanks!

examples/agent_dispatch.py
```diff
@@ -22,8 +22,8 @@
 """
 
 
-async def create_explicit_disptach(http_session: aiohttp.ClientSession):
-    agent_disptach_service = AgentDispatchService(
+async def create_explicit_dispatch(http_session: aiohttp.ClientSession):
+    agent_dispatch_service = AgentDispatchService(
```
davidzhao, line 26: @s-hamdananwar, could you update this example to use api.LiveKitAPI() as seen on our docs (also fixing the imports to import protobuf types from livekit.api instead of livekit.protocol.x

s-hamdananwar: @davidzhao I have added it in this PR https://github.com/livekit/python-sdks/pull/322

PR #322, opened 2024-12-10
fix agent dispatch example imports

s-hamdananwar:


examples/agent_dispatch.py
```diff
@@ -22,19 +13,17 @@
...
+        api.CreateAgentDispatchRequest(
+            agent_name=agent_name, room=room_name, metadata="my_metadata"
+        )
     )
-    dispatch = await agent_dispatch_service.create_dispatch(dispatch_request)
     print("created dispatch", dispatch)
-    dispatches = await agent_dispatch_service.list_dispatch(room_name=room_name)
+
+    dispatches = await lkapi.agent_dispatch.list_dispatch(room_name=room_name)
     print(f"there are {len(dispatches)} dispatches in {room_name}")
```
davidzhao, line 27: lg, btw we need a `await lkapi.aclose()` at the end to close out the session

s-hamdananwar: just added, thanks for the catch!

PR #323, opened 2024-12-11
Update publish_hue to publish at 30fps

davidzhao:


examples/publish_hue.py
```diff
@@ -38,8 +39,14 @@ async def main(room: rtc.Room):
     # publish a track
     source = rtc.VideoSource(WIDTH, HEIGHT)
     track = rtc.LocalVideoTrack.create_video_track("hue", source)
-    options = rtc.TrackPublishOptions()
-    options.source = rtc.TrackSource.SOURCE_CAMERA
+    options = rtc.TrackPublishOptions(
+        source=rtc.TrackSource.SOURCE_CAMERA,
+        simulcast=True,
+        video_encoding=rtc.VideoEncoding(
+            max_framerate=FPS,
```
nbsp, line 46: max_framerate might need to be a float here

theomonnom (approved):

examples/publish_hue.py
davidzhao: we should convert automatically? because that's non-obvious to a user

theomonnom: python should do it automatically?

PR #324, opened 2024-12-11
feat: add AVSynchronizer and examples

longcw:
moved from https://github.com/livekit/agents/pull/1168

### DIFF
--- a/examples/video-stream/audio_wave.py
+++ b/examples/video-stream/audio_wave.py
@@ -1,0 +1,320 @@
+import asyncio
+import logging
+import os
+import signal
+import time
+from collections import deque
+from dataclasses import dataclass
+from typing import AsyncIterable, Optional, Union
+
+import numpy as np
+from livekit import rtc, api
+
+try:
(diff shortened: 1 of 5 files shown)


examples/video-stream/audio_wave.py
```diff
@@ -0,0 +1,320 @@
...
+
+async def main(room: rtc.Room):
+    token = (
+        api.AccessToken()
+        .with_identity("python-publisher")
+        .with_name("Python Publisher")
+        .with_grants(
+            api.VideoGrants(
+                room_join=True,
+                room="room-ysBA-Q0hM",
```
davidzhao, line 229: nit, hardcoding a simpler room name would make it easier to test

davidzhao (approved):
🚀

examples/video-stream/audio_wave.py
longcw: oh yeah, I tested it using the agent playground and forgot to change it back a simple name.

PR #336, opened 2024-12-27
Add `close` function to Audio|Video Source

typester:


davidzhao (approved):
lg!

typester:
After reading [Lukas's comment](https://github.com/livekit/node-sdks/pull/365#discussion_r1898892649), I was also unsure during implementation whether the Python side should be made async, so I’m leaving this as an open question.

Should we await `self._join_handle` or `self._join_fut` ? If the answer is yes, I think it would be better to implement this as an `async aclose` and include the logic to wait for them. What do you think? @davidzhao @theomonnom

theomonnom:
Yes please, let's create a aclose function instead, we don't need to await the join handle but it is great to keep it async to be future proof on future changes we can do

typester:
Ok, I made it async!

davidzhao (approved):
👍

theomonnom (approved):

PR #338, opened 2024-12-30
fix av sync example

longcw:
Improved the video play example:
1. log the timestamps of the video and audio frames
2. add reset for av sync after the video finished


davidzhao (approved):
this version seems to resolve the sync issues that I had. lg.

PR #347, opened 2025-01-15
Add data stream support

lukasIO:


livekit-rtc/livekit/rtc/data_stream.py
```diff
@@ -0,0 +1,355 @@
...
+    def __init__(
+        self,
+        local_participant: LocalParticipant,
+        file_name: str,
+        topic: str = "",
+        extensions: Optional[Dict[str, str]] = {},
+        stream_id: str | None = None,
+        total_size: int | None = None,
+        mime_type: str = "",
+        destination_identities: List[str] = [],
```
longcw, line 314: it's better to not use mutable list or dict as default value.
> Default parameter values are evaluated when the function definition is executed. This means that the expression is evaluated once, when the function is defined, and that the same “pre-computed” value is used for each call. This is especially important to understand when a default parameter is a mutable object, such as a list or a dictionary: if the function modifies the object (e.g. by appending an item to a list), the default value is in effect modified. This is generally not what was intended. A way around this is to use None as the default, and explicitly test for it in the body of the function (https://docs.python.org/2/reference/compound_stmts.html#function)
```
destination_identities: Optional[List[str]] = None

...
# inside function
destination_identities = destination_identities or []

File: livekit-rtc/livekit/rtc/data_stream.py [hunk 2]
```diff
@@ -0,0 +1,355 @@
...
+                lambda e: e.send_stream_trailer.async_id
+                == resp.send_stream_trailer.async_id
+            )
+        finally:
+            FfiClient.instance.queue.unsubscribe(queue)
+
+        if cb.send_stream_chunk.error:
+            raise ConnectionError(cb.send_stream_trailer.error)
+
+    async def close(self):
```
davidzhao, line 244:
```suggestion
    async def aclose(self):
```

livekit-rtc/livekit/rtc/data_stream.py
```diff
@@ -0,0 +1,355 @@
...
+class BaseStreamWriter:
+    def __init__(
+        self,
+        local_participant: LocalParticipant,
+        topic: str = "",
+        extensions: Optional[Dict[str, str]] = {},
+        stream_id: str | None = None,
+        total_size: int | None = None,
+        mime_type: str = "",
+        destination_identities: List[str] = [],
```
davidzhao, line 159: would like to make these kwargs only

```suggestion
        *,
        topic: str = "",
        extensions: Optional[Dict[str, str]] = {},
        stream_id: str | None = None,
        total_size: int | None = None,
        mime_type: str = "",
        destination_identities: List[str] = [],
```

livekit-rtc/livekit/rtc/data_stream.py
```diff
@@ -0,0 +1,355 @@
...
+class TextStreamWriter(BaseStreamWriter):
+    def __init__(
+        self,
+        local_participant: LocalParticipant,
+        topic: str = "",
+        extensions: Optional[Dict[str, str]] = {},
+        stream_id: str | None = None,
+        total_size: int | None = None,
+        reply_to_id: str | None = None,
+        destination_identities: List[str] = [],
```
davidzhao, line 261: same as above

davidzhao (commented):
really great work!

livekit-rtc/livekit/rtc/data_stream.py
lukasIO: done!

livekit-rtc/livekit/rtc/data_stream.py
lukasIO: done

livekit-rtc/livekit/rtc/data_stream.py
```diff
@@ -0,0 +1,362 @@
...
+            attributes=dict(header.attributes),
+            name=header.byte_header.name,
+        )
+        self._queue: RingQueue[proto_DataStream.Chunk | None] = RingQueue(capacity)
+
+    def _on_chunk_update(self, chunk: proto_DataStream.Chunk):
+        self._queue.put(chunk)
+
+    def _on_stream_close(self, trailer: proto_DataStream.Trailer):
+        self._queue.put(None)
```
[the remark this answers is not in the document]
davidzhao: I think we'd need to remember that we've received close.. and during __anext__, when the queue is empty we can return

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -539,6 +549,119 @@ async def set_attributes(self, attributes: dict[str, str]) -> None:
...
+
+        await writer._send_header()
+
+        return writer
+
+    async def send_file(
+        self,
+        file_path: str,
+        topic: str = "",
+        destination_identities: Optional[List[str]] = None,
```
davidzhao, line 636: would be great to add `file_name` as an optional kwarg.

picking a default filename is great.. but it's also helpful if I want a local tempfile to be sent, and don't know/care about the actual filename on disk

davidzhao (commented):
nice! just a few minor things

livekit-rtc/livekit/rtc/data_stream.py
```diff
@@ -0,0 +1,362 @@
...
+        self._info = ByteStreamInfo(
+            stream_id=header.stream_id,
+            mime_type=header.mime_type,
+            topic=header.topic,
+            timestamp=header.timestamp,
+            size=header.total_length,
+            attributes=dict(header.attributes),
+            name=header.byte_header.name,
+        )
+        self._queue: RingQueue[proto_DataStream.Chunk | None] = RingQueue(capacity)
```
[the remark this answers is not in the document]
lukasIO: this is how we also do it in other places in the python SDK, will leave it with | None for consistency

livekit-rtc/livekit/rtc/participant.py
lukasIO: optional `name` is fine with me, but considering the fact that we now went from naming it `file_name` to `name` on the receiving side, I don't think we should call it `file_name` here either, wdyt?

livekit-rtc/livekit/rtc/data_stream.py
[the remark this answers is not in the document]
lukasIO: I think the queue is unbounded, so this shouldn't happen. Consulted @theomonnom and this implementation should be in line with how we handle audio_stream [here](https://github.com/livekit/python-sdks/blob/lukas/data-streams/livekit-rtc/livekit/rtc/audio_stream.py#L224-L231).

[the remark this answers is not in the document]
lukasIO: ok, updated it to use asyncio.Queue, the annoying part is that the `put` method is async, so we end up with room event handling code that looks like
```
elif which == "stream_chunk_received":
            asyncio.run(self._handle_stream_chunk(event.stream_chunk_received.chunk))
```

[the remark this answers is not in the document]
davidzhao: > I think the queue is unbounded

I see. I just assumed it's bounded when I saw RingQueue

livekit-rtc/livekit/rtc/participant.py
davidzhao: sure!

davidzhao (approved):
🍾

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -539,6 +549,119 @@ async def set_attributes(self, attributes: dict[str, str]) -> None:
...
+
+        return writer.info
+
+    async def stream_bytes(
+        self,
+        name: str,
+        *,
+        total_size: int | None = None,
+        mime_type: str = "application/octet-stream",
+        extensions: Optional[Dict[str, str]] = None,
```
theomonnom, line 608: Let's do like the send_text method
`extensions: dict[str, str]`

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -539,6 +549,119 @@ async def set_attributes(self, attributes: dict[str, str]) -> None:
...
+        stream_id: str | None = None,
+        destination_identities: Optional[List[str]] = None,
```
theomonnom, line 610: destination_identities: List[str] = [],

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -539,6 +549,119 @@ async def set_attributes(self, attributes: dict[str, str]) -> None:
...
+            topic=topic,
+        )
+
+        await writer._send_header()
+
+        return writer
+
+    async def send_file(
+        self,
+        file_path: str,
```
theomonnom, line 634: I would use kwargs here

livekit-rtc/livekit/rtc/participant.py
lukasIO: I specifically changed this to optional because @longcw had concerns about defaulting a list to an empty list in method parameters

lukasIO: see https://github.com/livekit/python-sdks/pull/347#discussion_r1920930040

livekit-rtc/livekit/rtc/data_stream.py
```diff
@@ -0,0 +1,365 @@
...
+
+if TYPE_CHECKING:
+    from .participant import LocalParticipant
+
+
+STREAM_CHUNK_SIZE = 15_000
+
+
+@dataclass
+class BaseStreamInfo(TypedDict):
```
longcw, line 38: This should be either a dataclass or a TypedDict, but not both. I prefer to make it a dataclass.

lukasIO: thanks! updated in #352

PR #348, opened 2025-01-17
add pydantic support to rtc.VideoFrame & rtc.AudioFrame

theomonnom:
- this package still doesn't depend on pydantic 
- data is serialized in b64


davidzhao (approved):
lg, but is the rust lib update intentional?

PR #350, opened 2025-01-27
feat: audio filter plugin

typester:
The `rust-sdk` implementation (https://github.com/livekit/rust-sdks/pull/559) must be merged before this pull request.


livekit-rtc/livekit/rtc/_ffi_client.py
```diff
@@ -212,7 +212,7 @@ def __init__(self) -> None:
         self._queue = FfiQueue[proto_ffi.FfiEvent]()
 
         ffi_lib.livekit_ffi_initialize(
-            ffi_event_callback, True, b"python", __version__.encode("ascii")
+            ffi_event_callback, False, b"python", __version__.encode("ascii")
```
bcherry, line 215: is there a reason `capture_logs` was changed?

bcherry (commented):
lgtm, lmk when it's ready for final review

livekit-rtc/livekit/rtc/_ffi_client.py
typester: Ah, that was my bad. I set the value for debugging and accidentally committed it. Will remove, thank you!

typester: Fixed in [dd2e973](https://github.com/livekit/python-sdks/pull/350/commits/dd2e9732a95dff4f1ba4f3ed07b9d332da52e451)

livekit-rtc/livekit/rtc/audio_stream.py
```diff
@@ -54,6 +55,8 @@ def __init__(
         capacity: int = 0,
         sample_rate: int = 48000,
         num_channels: int = 1,
+        enable_filter: Any = None,
+        filter_options: dict[str, Any] | None = None,
```
theomonnom, line 59: I would be more eager to add strict typings here.
I think we should remove the filter_options and put it inside the class.
I was imagining something like:

`AudioStream(krisp=KrispFilter())`

I'm not sure why we make it a generic interface, feels easier to explicitly mention Krisp, wdyt?

livekit-rtc/livekit/rtc/audio_filter.py
```diff
@@ -0,0 +1,32 @@
...
+        req = proto_ffi.FfiRequest()
+        req.load_audio_filter_plugin.plugin_path = path
+        req.load_audio_filter_plugin.dependencies[:] = (
+            dependencies if dependencies is not None else []
+        )
+        req.load_audio_filter_plugin.options = json.dumps(
+            {
+                "url": url,
+                "token": token,
+            }
```
theomonnom, line 23: Should we make it strictly typed inside our protobufs definition instead of using json? The auth function could accept those as arguments (no need for a prost dependency)

livekit-rtc/livekit/rtc/room.py
```diff
@@ -126,7 +128,11 @@ def __init__(self, message: str):
 
 
 class Room(EventEmitter[EventTypes]):
-    def __init__(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> None:
+    def __init__(
+        self,
+        loop: Optional[asyncio.AbstractEventLoop] = None,
+        filters: Optional[List[Any]] = None,
```
theomonnom, line 134: Using the constructor is tricky because users don't construct it on agents. IMO we should only expose this API inside the rtc.AudioStream for now

livekit-rtc/livekit/rtc/audio_stream.py
typester: Since I'm not familiar with Python, any suggestion like this is super welcome! Let me think about this

livekit-rtc/livekit/rtc/room.py
typester: Yeah, I think so. But the Krisp filter needs to be initialized with the LIVEKIT_URL and TOKEN parameters to verify the token. So, I've initialized it here.

Instead, do you mean that it would be better to pass all of this information in the AudioStream constructor, as you suggested earlier? Or creating a separate `init` function to pass these info? Something like:

```
krisp.init(url, token)
```

or something like that. I think the latter is better

theomonnom: no strong opinion, tho I know dz wanted to avoid the auth on the python side (so grabbing the url/token on the rust side)

livekit-rtc/livekit/rtc/audio_stream.py
```diff
@@ -54,6 +55,7 @@ def __init__(
         capacity: int = 0,
         sample_rate: int = 48000,
         num_channels: int = 1,
+        audio_filter: Optional[Tuple[str, dict[str, any]]] = None,
```
theomonnom, line 58: Any way to get something that looks more like https://github.com/livekit/python-sdks/pull/350#discussion_r1945572208
This type isn't self-explanatory

typester: I intended to cover this in the filter module side, because the option parameter will vary depending on the filter module. It looks something like this:

```
audio_filter=noise_filter.NC()
```

Please refer to the noise_filter repo for more details

theomonnom: I'm not sure why we want to make this generic, if we have more filters in the future we could also add new keyword arguments? I propose to change the argument name to be directly krisp

I also doubt we're going to have more filters?

typester: Exposing the raw API is better in terms of architecture, and I agreed with this in our previous discussion. But it ended up with it's not good with our metric system, wasn't it? So we decided to extend AudioStream for now.

In that case, I would say that it would be better to handle the filter options in noise_filter side and to keep AudioStream  generic. This would make updates to noise_filter easier. What do you think?

theomonnom: Sorry, I mean the echo cancellation in the previous message. 

What I mean is why we want to keep the name generic like audio_filter instead of krisp
`krisp=KrispFilter(options)`

typester: Oh I see. I don't have a strong opinion on this. We're creating a noise filter package as `livekit-noise-filter` (Initially, it was named `livekit-krisp`, but we renamed it.) So, would it be the latter?

I introduced two parameters to the python-sdk.

The first one is for `RoopOptions`: (This is required to register the filter to the room)

```python
await room.connect(
    os.getenv("LIVEKIT_URL"), token,
    rtc.RoomOptions(
        auto_subscribe=True,
        audio_filters=[noise_filter],
    )
)
```

The second is for `AudioStream`:

```python
stream = rtc.AudioStream.from_track(
    track=track,
    audio_filter=noise_filter.NC(),
)
```

So, should both options be named `noise_cancellation_modules`  and `noise_cancellation`? How do they sound?

theomonnom: Yes, I would name them krisp_filter. To add to that, do we really need to add this requirement to add the filters inside the RoomOptions? Can registering be skipped and happen implicitly?

theomonnom: I think if we follow the agents stuff, it should be a livekit-plugins-krisp package

theomonnom: If it isn't that much work to do, I think it is worth it. So people only have to worry about one place in their code to enable Krisp.

The plugin would need to be contributed inside the agents repository, we have a livekit-plugins-minimal example [here](https://github.com/livekit/agents/tree/main/livekit-plugins/livekit-plugins-minimal).

typester: > If it isn't that much work to do, I think it is worth it. This way, people only need to focus on a single spot in their code to enable Krisp.
Agreed. Will try to make it.

For the plugin repository, is it possible to implement a plugin in a separate repository? The krisp filter plugin needs to be bundled with non-opensource rust-ffi. so I don't think it's a good idea to put it in the plugin repo, that likely needs to implement some tricky CIs

davidzhao: yeah I think the plugin should exist in the krisp filters repo, since it's not the same license & closed sourced. we just have to publish a package that's `livekit-plugins-krisp`, and that should be enough.

typester: ok, I did the following changes:

- remove audio_filter parameters from RoomOptions, now filters are automatically registered.
- rename audio_filter parameter of rtc.NativeStream parameter to krisp.

So now we can do:

```python
from livekit.plugins import krisp

# ...

stream = rtc.AudioStream.from_participant(
    participant=participant,
    track_source=rtc.TrackSource.SOURCE_MICROPHONE,
    krisp=krisp.BVC(),
 )
```

Or something like that.

livekit-rtc/livekit/rtc/audio_stream.py
```diff
@@ -54,6 +55,7 @@ def __init__(
         capacity: int = 0,
         sample_rate: int = 48000,
         num_channels: int = 1,
+        noise_cancellation: Optional[Tuple[str, dict[str, Any]]] = None,
```
bcherry, line 58: I'm not sure what the resolution in the other comment thread was but I'd suggest at least adding a type alias for this e.g. `NoiseCancellationOptions` so it's clear

also please add documentation to the docstring below

livekit-rtc/livekit/rtc/audio_stream.py
```diff
@@ -112,6 +119,7 @@ def from_participant(
...
```
bcherry, line 122: same here, please document the new parameter

livekit-rtc/livekit/rtc/audio_stream.py
```diff
@@ -155,6 +164,7 @@ def from_track(
...
```
bcherry, line 167: same here, please document the new parameter

livekit-rtc/livekit/rtc/audio_stream.py
theomonnom: I think a dataclass with the available options would be better

typester: Added NoiseCancellationOptions as a dataclass and exported it for using it by noise-cancellation module.

Also a brief documentation was added!

livekit-rtc/livekit/rtc/audio_stream.py
```diff
@@ -39,6 +40,12 @@ class AudioFrameEvent:
     frame: AudioFrame
 
 
+@dataclass
+class NoiseCancellationOptions:
+    module_id: str
+    options: dict[str, Any]
```
theomonnom, line 46: What should the options include? This is the final unclear point.

theomonnom (approved):
otherwise lgtm!! 🚀

livekit-rtc/livekit/rtc/room.py
```diff
@@ -852,6 +860,12 @@ def _create_remote_participant(
         self._remote_participants[participant.identity] = participant
         return participant
 
+    def _filter_handle(self, plugin) -> int | None:
```
theomonnom, line 863: This seems unused? Or maybe it is on the krisp-repo?

typester: Oh, this is unused now. It was used for associate room and filters, but it was moved to rust-sdk. I'll remove this.

livekit-rtc/livekit/rtc/audio_stream.py
typester: The options are eventually encoded to JSON and will be passed to the FFI as Cstr.

The noise-cancellation module should have functions like:

```
def BVC() -> rtc.NoiseCancellationOptions:
```

And users this this type via the function like:

```
from livekit.plugins import noise_cancellation
# ...

rtc.AudioStream(
  # ...
  noise_cancellation=noise_cancellation.BVC(),
)
```

So users should not create this Options by their hand.

bcherry (approved):

PR #352, opened 2025-01-29
Rework data stream API

lukasIO:
addresses feedback from @theomonnom on https://github.com/livekit/python-sdks/pull/347 after it got merged

### DIFF
--- a/livekit-rtc/livekit/rtc/room.py
+++ b/livekit-rtc/livekit/rtc/room.py
@@ -437,7 +437,7 @@ class Room(EventEmitter[EventTypes]):
             return
 
         await self._drain_rpc_invocation_tasks()
-        await self._drain_data_stream__tasks()
+        await self._drain_data_stream_tasks()
 
         req = proto_ffi.FfiRequest()
         req.disconnect.room_handle = self._ffi_handle.handle  # type: ignore
@@ -478,7 +478,7 @@ class Room(EventEmitter[EventTypes]):
 
         # Clean up any pending RPC invocation tasks
         await self._drain_rpc_invocation_tasks()
-        await self._drain_data_stream__tasks()
+        await self._drain_data_stream_tasks()
 
     def _on_rpc_method_invocation(self, rpc_invocation: RpcMethodInvocationEvent):
         if self._local_participant is None:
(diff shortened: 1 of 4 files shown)


livekit-rtc/livekit/rtc/room.py
```diff
@@ -747,9 +747,11 @@ def _on_room_event(self, event: proto_room.RoomEvent):
                 event.stream_header_received.participant_identity,
             )
         elif which == "stream_chunk_received":
-            asyncio.gather(self._handle_stream_chunk(event.stream_chunk_received.chunk))
+            asyncio.create_task(
```
lukasIO, line 750: @theomonnom in general this became an issue when switching from the RingQueue to asyncio.Queue as `put` is async with it. Maybe we should avoid this and switch back to RingQueue after all?

theomonnom: We should keep a reference to every task that has been created.
https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

Otherwise it may no be executed completely.
Let's do the same implementation as what Ben did for RPC (This should also be cancelled on room disconnect I guess)

livekit-rtc/livekit/rtc/room.py
```diff
@@ -835,6 +818,41 @@ async def _handle_stream_trailer(self, trailer: proto_room.DataStream.Trailer):
             await file_reader._on_stream_close(trailer)
             self._byte_stream_readers.pop(trailer.stream_id)
 
+    async def _drain_rpc_invocation_tasks(self) -> None:
+        if self._rpc_invocation_tasks:
+            for task in self._rpc_invocation_tasks:
+                task.cancel()
+            await asyncio.gather(*self._rpc_invocation_tasks, return_exceptions=True)
+
+    async def _drain_data_stream__tasks(self) -> None:
```
davidzhao, line 827: nit:

```suggestion
    async def _drain_data_stream_tasks(self) -> None:
```

davidzhao (approved):

livekit-rtc/livekit/rtc/room.py
longcw, line 1: should this warning be a debug or removed: https://github.com/livekit/python-sdks/pull/352/files#diff-e6d6a673948fca293a54c3930f2c95f3fec05cc33b6dc8c27935584da4529b17L802

lukasIO: uff, thanks, that was a debug left over, removed

PR #357, opened 2025-02-05
version bump

davidzhao:


livekit-rtc/livekit/rtc/version.py
```diff
@@ -1 +1 @@
-__version__ = "0.19.1"
+__version__ = "0.19.2"
```
bcherry, line 1: If we're planning to release everything in main then we should do a minor version increase instead to 0.20.0 since this release includes data streams.

theomonnom:
@bcherry can we bump rust-sdks to include the rpc race fix

theomonnom (approved):
seems like this is done here https://github.com/livekit/python-sdks/pull/356
lgtm then

PR #361, opened 2025-02-14
Add auto chunking for text stream writer

lukasIO:
- removes mid-stream chunk overrides
- this allows for `writer.write(text)` to perform auto chunking, mirroring the same behaviour as on byte streams
- also allows for send_text to be a thin auto-closing wrapper over stream_text
- allows for `text_reader` to be a simple async iterable of type `str` instead of the `TextStreamUpdate` data class (again improved consistency with how byte stream reader works)


davidzhao (approved):
lgtm

PR #369, opened 2025-02-25
check data stream closed

longcw:


livekit-rtc/livekit/rtc/data_stream.py
```diff
@@ -235,6 +240,8 @@ async def _send_trailer(self, trailer: proto_DataStream.Trailer):
         if cb.send_stream_chunk.error:
             raise ConnectionError(cb.send_stream_trailer.error)
 
+        self._closed = True
```
lukasIO, line 243: I think we should set self._closed = True already in the `aclose` method before awaiting sending of the trailer

lukasIO: and therefore also move the self._closed check into the aclose method

longcw: yeah, makes sense

PR #395, opened 2025-03-13
WIP: Move RPC handlers to room

typester:
Rust-side: https://github.com/livekit/rust-sdks/pull/599


livekit-rtc/livekit/rtc/room.py
```diff
@@ -483,7 +564,7 @@ def _on_rpc_method_invocation(self, rpc_invocation: RpcMethodInvocationEvent):
 
         if rpc_invocation.local_participant_handle == self._local_participant._ffi_handle.handle:
```
typester, line 565: Is this local_participant check still needed?

bcherry: i think this is here in the case that you connect to multiple rooms simultaneously?

examples/rpc_deprecated.py
davidzhao, line 1: we should get rid of this one, don't want to show users using it the incorrect way

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -331,6 +335,8 @@ def register_rpc_method(
         handler: Optional[Callable[[RpcInvocationData], Union[Awaitable[str], str]]] = None,
     ) -> Union[None, Callable]:
         """
+        Deprecated
```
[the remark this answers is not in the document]
davidzhao: perhaps using this library? https://github.com/laurent-laporte-pro/deprecated

this will be part of Python 3.13. the implementation seems simple enough: https://stackoverflow.com/questions/2536307/decorators-in-the-python-standard-lib-deprecated-specifically

davidzhao (commented):
lg, just a couple of nits

examples/rpc_deprecated.py
typester: Removed in b772e5d79d5a0bb6dc14c345076bab20eb943372

livekit-rtc/livekit/rtc/participant.py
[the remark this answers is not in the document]
typester: Added in 04e5e0644a6143aca8a639c5cca5b4251db4a80b

@davidzhao Could you please check if I’ve added the dependency correctly? I added it in `setup.py`, but I’m not sure if that’s the right place.

davidzhao (approved):
lgtm

bcherry (approved):

livekit-rtc/setup.py
```diff
@@ -58,7 +58,12 @@ def finalize_options(self):
     license="Apache-2.0",
     packages=setuptools.find_namespace_packages(include=["livekit.*"]),
     python_requires=">=3.9.0",
-    install_requires=["protobuf>=4.25.0", "types-protobuf>=3", "aiofiles>=24"],
+    install_requires=[
+        "protobuf>=4.25.0",
+        "types-protobuf>=3",
+        "aiofiles>=24",
+        "deprecated>=1.2.18",
```
theomonnom, line 65: do we want to add a new package just for deprecation?

let's just log a warning no?

theomonnom: just saw this comment, either way https://github.com/livekit/python-sdks/pull/395#discussion_r1994960391 ...

davidzhao: we can log a warning too! I don't have a preference.. the only preference is making sure users know they are using a deprecated function

PR #429, opened 2025-04-29
expose rtc session stats 

theomonnom:


theomonnom:
example here: https://github.com/livekit/agents/pull/2155/files

Issue #430, opened 2025-05-02
Bug: ByteStreamWriter sends first chunk with index 1, causing client errors


5Hyeons:
**Description:**
There is an inconsistency in chunk indexing between `ByteStreamWriter` and `TextStreamWriter` in livekit/rtc/data_stream.py. The `ByteStreamWriter.write` method incorrectly increments the `_next_chunk_index` before using it for the chunk_index field in the `proto_DataStream.Chunk` message. **_This results in the first chunk being sent with index 1._**

**Code References:**
Incorrect ByteStreamWriter logic: https://github.com/livekit/python-sdks/blob/ea42eb8a77ee2954efb002687b7e04b03c3f79ab/livekit-rtc/livekit/rtc/data_stream.py#L342-L348
Correct TextStreamWriter logic: https://github.com/livekit/python-sdks/blob/ea42eb8a77ee2954efb002687b7e04b03c3f79ab/livekit-rtc/livekit/rtc/data_stream.py#L286-L294

**Problem:**
Most stream consumers, including the LiveKit Unity SDK (`ByteStreamReader`), expect the first chunk index of a stream to be 0. 
When receiving a chunk with index 1 without seeing index 0 first, the client raises an error, typically like "expected chunk index to be exactly one more than the previous". This prevents byte streams from working correctly with such clients.

Text streams work correctly because `TextStreamWriter` uses the current `_next_chunk_index` _(which starts at 0)_ before incrementing it.

**Proposed Solution:**
Modify `ByteStreamWriter.write` to use `_next_chunk_index` for the `chunk_index` before incrementing it, consistent with `TextStreamWriter`.

PR #431, opened 2025-05-02
Fix: Start ByteStreamWriter chunk index at 0 #430

5Hyeons:
This PR fixes a bug in `ByteStreamWriter` where the first data chunk was incorrectly sent with **_chunk_index = 1_** instead of the expected 0.

**Problem:**
As described in Issue #430 , the `_next_chunk_index` was incremented before being used in the write method, causing incompatibility with clients (like the Unity SDK's ByteStreamReader) that expect stream chunk indices to start at 0. 
This differs from the behavior of `TextStreamWriter`, which correctly starts at index 0.

**Changes:**
The ByteStreamWriter.write method has been modified to:
Use the current value of `_next_chunk_index` (which starts at 0 in `BaseStreamWriter`) for the chunk_index.
Increment `_next_chunk_index` after the chunk message has been sent via `_send_chunk`.
This aligns the behavior with `TextStreamWriter` and ensures correct 0-based indexing for byte streams, resolving client-side errors related to chunk index expectation.

### DIFF
--- a/livekit-rtc/livekit/rtc/data_stream.py
+++ b/livekit-rtc/livekit/rtc/data_stream.py
@@ -340,13 +340,13 @@ class ByteStreamWriter(BaseStreamWriter):
             ]
 
             for chunk in chunked_data:
-                self._next_chunk_index += 1
                 chunk_msg = proto_DataStream.Chunk(
                     stream_id=self._header.stream_id,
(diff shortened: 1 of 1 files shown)


lukasIO (approved):
thanks for catching that!

PR #456, opened 2025-06-16
add user_data to AudioFrame

longcw:


livekit-rtc/livekit/rtc/audio_frame.py
```diff
@@ -100,6 +101,13 @@ def _proto_info(self) -> proto_audio.AudioFrameBufferInfo:
         audio_info.samples_per_channel = self.samples_per_channel
         return audio_info
 
+    @property
+    def user_data(self) -> dict[str, Any]:
```
theomonnom: We don't use the underscore on other parts of our code. So just for consistency 
```suggestion
    def userdata(self) -> dict[str, Any]:
```

PR #457, opened 2025-06-16
wip: handle room updates

davidzhao:


livekit-rtc/livekit/rtc/room.py
```diff
@@ -235,6 +238,62 @@ def e2ee_manager(self) -> E2EEManager:
...
         return self._e2ee_manager
 
+    @property
+    def num_participants(self) -> int:
+        """Gets the number of participants in the room.
+
+        Returns:
+            int: The number of participants in the room.
+        """
+        return self._info.num_participants
```
theomonnom, line 249: Are we sure this is in sync with `len(room.participants)`
nothing guarantees it and it is quite confusing IMO

davidzhao: it's not in sync.. but eventually consistent. in our clients, we are returning this value because of certain modes (livestream mode), viewers aren't sent back to every other user in the room

PR #461, opened 2025-06-24
Readme: added badge for livekit-api

RSATom:
I surprised `livekit` and `livekit-api` have different versioning. And I think it can be useful to see versions of both in Readme to avoid confusions for others.


davidzhao (approved):
good call. thanks!

PR #473, opened 2025-07-23
make verify_signature optional for TokenVerifier

longcw:


livekit-api/livekit/api/access_token.py
```diff
@@ -214,7 +214,10 @@ def verify(self, token: str) -> Claims:
             algorithms=["HS256"],
             leeway=self._leeway.total_seconds(),
         )
+        return self.decode_claims(claims)
 
+    @staticmethod
+    def decode_claims(claims: dict[str, Any]) -> Claims:
```
davidzhao: instead of taking in a dict, it'd be better to go straight from token -> claims, while including a parameter to skip the secret verification. it seems awkward to duplicate the jwt decode operation in agents

longcw: updated

PR #486, opened 2025-08-19
fixed calls from 'set' to 'register'

aryeila:
Fix for error in data streams examples for byte & text stream handlers.


bcherry (approved):
thank you!

PR #493, opened 2025-09-09
Media Devices

chenosaurus:
- Add a new class `MediaDevices` that provides centralized access to local audio input/outputs.
- Enables AEC loop when using both mic & speaker.

### DIFF
--- a/livekit-rtc/livekit/rtc/media_devices.py
+++ b/livekit-rtc/livekit/rtc/media_devices.py
@@ -1,0 +1,448 @@
+# Copyright 2025 LiveKit, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import asyncio
+from dataclasses import dataclass
+from typing import Any, AsyncIterator, Optional
+
+import numpy as np
+import sounddevice as sd
+import threading
+
+from . import AudioSource
+from .audio_frame import AudioFrame
+from .apm import AudioProcessingModule
+
+"""
+Media device helpers built on top of the `sounddevice` library.
+
+This module provides a small, Pythonic helper around native audio I/O for
+LiveKit RTC usage:
+
+- Capture the default microphone and feed frames into `rtc.AudioSource`.
+- Optionally enable audio processing via `rtc.AudioProcessingModule` (AEC,
+  noise suppression, high-pass filter, AGC). Frames are processed in 10 ms
+  chunks as required by APM.
+- Play arbitrary audio frames to the default speaker. When AEC is enabled on
+  the microphone, the `OutputPlayer` can feed the APM reverse stream so echo
+  cancellation has access to render (speaker) audio.
+
+Notes on AEC wiring:
+- AEC requires feeding both capture (mic) and reverse (speaker) paths into
+  the same APM instance. This module does not automatically capture output from
+  other players. To enable AEC, the output player feeds APM's reverse stream
+  and we set stream delays derived from PortAudio timing.
+"""
+
+
+DEFAULT_SAMPLE_RATE = 48000
+DEFAULT_CHANNELS = 1
+FRAME_SAMPLES = 480  # 10 ms at 48 kHz
+BLOCKSIZE = 4800  # 100 ms I/O buffer size for sounddevice
+
+
+def _ensure_loop(loop: Optional[asyncio.AbstractEventLoop]) -> asyncio.AbstractEventLoop:
+    return loop or asyncio.get_event_loop()
+
+
+class _APMDelayEstimator:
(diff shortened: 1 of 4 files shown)


livekit-rtc/livekit/rtc/audio_mixer.py
```diff
@@ -184,6 +184,10 @@ async def _get_contribution(
             except StopAsyncIteration:
                 exhausted = True
                 break
+            # AudioStream may yield either AudioFrame or AudioFrameEvent; unwrap if needed
```
theomonnom: The correct typing should be:

`self, stream: AsyncIterator[AudioFrame | AudioFrameEvent], buf: np.ndarray`

livekit-rtc/livekit/rtc/media_devices.py
```diff
@@ -0,0 +1,478 @@
...
+            "samplerate": self._in_sr,
+            "blocksize": self._blocksize,
+        }
+        try:
+            init_params = inspect.signature(sd.InputStream.__init__).parameters
+            if "mapping" in init_params and mapping is not None:
+                stream_kwargs["mapping"] = mapping
+            elif mapping is not None:
+                logging.getLogger(__name__).warning(
+                    "sounddevice.InputStream does not support 'mapping' in this version; "
```
theomonnom: Do we have to support that? or only latest?

livekit-rtc/livekit/rtc/media_devices.py
```diff
@@ -0,0 +1,478 @@
...
+        except Exception:
+            # If inspection fails for any reason, fall back without mapping
+            if mapping is not None:
+                logging.getLogger(__name__).warning(
+                    "Unable to inspect sounddevice.InputStream.__init__; "
+                    "ignoring input_channel_index=%s",
+                    input_channel_index,
+                )
+
+        input_stream = sd.InputStream(**stream_kwargs)
```
theomonnom: nit: Ideally we keep the typed args (no dict)

examples/local_audio/full_duplex.py
```diff
@@ -0,0 +1,147 @@
...
+        if pub_sid is not None:
+            streams_by_pub.pop(pub_sid, None)
+
+    def on_track_subscribed(
+        track: rtc.Track,
+        publication: rtc.RemoteTrackPublication,
+        participant: rtc.RemoteParticipant,
+    ):
+        if track.kind == rtc.TrackKind.KIND_AUDIO:
+            stream = rtc.AudioStream(track, sample_rate=48000, num_channels=1)
```
theomonnom: I think we should abstract further down, and directly allow to add a AudioTrack

livekit-rtc/livekit/rtc/media_devices.py
```diff
@@ -0,0 +1,478 @@
...
+            input_stream=input_stream,
+            task=task,
+            apm=apm,
+            delay_estimator=delay_estimator,
+        )
+
+    def open_output(
+        self,
+        *,
+        apm_for_reverse: Optional[AudioProcessingModule] = None,
```
theomonnom: This should be exposed to the user. Let's make this an internal detail

theomonnom (commented):
Nice!!

livekit-rtc/livekit/rtc/media_devices.py
chenosaurus: do you mean should not?

theomonnom: Yes sry, it shouldn't

examples/local_audio/full_duplex.py
chenosaurus: do you mean we should add a `add_track` and `remove_track` method to the AudioMixer class?

theomonnom: yeah

theomonnom: But maybe more like player.add_track

chenosaurus: hmm yea that would make sense to add to media_devices to avoid users having to deal w/ a mixer to just play tracks back.  Lemme add this.

livekit-rtc/livekit/rtc/media_devices.py
```diff
@@ -0,0 +1,581 @@
...
+# limitations under the License.
+
+from __future__ import annotations
+
+import asyncio
+from dataclasses import dataclass
+from typing import Any, AsyncIterator, Optional
+
+import numpy as np
+import sounddevice as sd  # type: ignore[import-not-found, import-untyped]
```
theomonnom: can we make sure to import sounddevice lazily

livekit-rtc/livekit/rtc/media_devices.py
```diff
@@ -0,0 +1,595 @@
...
+
+        # Create audio stream for this track
+        stream = AudioStream(track, sample_rate=self._sample_rate, num_channels=self._num_channels)
+        # Wrap the stream with an adapter to convert AudioFrameEvent to AudioFrame
+        stream_iterator = _AudioStreamIterator(stream)
+
+        self._track_streams[track.sid] = (stream, stream_iterator)
+        self._mixer.add_stream(stream_iterator)
+
+    async def remove_track(self, track: Track) -> None:
```
theomonnom, line 266: should this method be sync instead of async? like add_track

chenosaurus: it awaits `stream.aclose()` in that function

theomonnom: I’m thinking more in terms of API design rather than the underlying reasons. For consistency, I’d make both methods async

theomonnom (approved):

PR #494, opened 2025-09-09
feat: support targeted number and address updates for sip trunks

rektdeckard:
Ticket: https://linear.app/livekit/issue/TEL-162/update-python-sdk-to-support-sip-listupdate-operations

Adds support for targeted updates (add/del) for numbers and addresses on SIP inbound/outbound trunks.

```py
await update_inbound_trunk_fields(trunk_id, numbers=ListUpdate(add=["1234567890"]))
```

### DIFF
--- a/livekit-api/livekit/api/sip_service.py
+++ b/livekit-api/livekit/api/sip_service.py
@@ -148,9 +148,9 @@ class SipService(Service):
         self,
         trunk_id: str,
         *,
-        numbers: Optional[list[str]] = None,
-        allowed_addresses: Optional[list[str]] = None,
-        allowed_numbers: Optional[list[str]] = None,
+        numbers: Optional[ListUpdate | list[str]] = None,
+        allowed_addresses: Optional[ListUpdate | list[str]] = None,
+        allowed_numbers: Optional[ListUpdate | list[str]] = None,
         auth_username: Optional[str] = None,
         auth_password: Optional[str] = None,
         name: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


livekit-api/livekit/api/sip_service.py
```diff
@@ -167,11 +167,20 @@ async def update_inbound_trunk_fields(
...
+            if isinstance(numbers, ListUpdate):
+                update.numbers = numbers
+            else:
+                update.numbers.set.extend(numbers)
         if allowed_addresses is not None:
-            update.allowed_addresses.set.extend(allowed_addresses)
+            if isinstance(allowed_addresses, ListUpdate):
+                update.allowed_addresses = allowed_addresses
+            else:
+                update.allowed_addresses.set.extend(allowed_addresses)
```
nishadmusthafa, line 178: This didn't occur to me during our call Tobias. We would need to update the field depending on what is inside the [ListUpdate type](https://github.com/livekit/protocol/blob/main/protobufs/livekit_models.proto#L36). The current code works if it's a set. If it's an add, we would need to extend and if it's a del, we would need to remove items. If it's clear, then set it to an empty list.

rektdeckard: We're forwarding the whole ListUpdate, if that's what is passed as an argument. Not sure I follow?

nishadmusthafa: Ahh.. right. I confused it with the list. LGTM!

PR #496, opened 2025-09-11
fix: update protos and allow ListUpdate.{add,remove} in SIP cmds

rektdeckard:
Regenerate proto bindings and update SIP APIs to fully support `ListUpdate` type actions. In the [previous PR](https://github.com/livekit/python-sdks/pull/494), we lacked support for the `del` operation, since it is a reserved Python keyword and was not being generated in the Python protobuf bindings.


nishadmusthafa (approved):

rektdeckard:
@theomonnom tests seem to have trouble downloading `livekit-ffi v0.12.32`. Any idea why that might be?

PR #512, opened 2025-10-09
Add support for datachannel encryption

lukasIO:
also exposes participantEncryptionChanged event and updates rust sdks (currently not to a new FFI build, which needs to get fixed)

### DIFF
--- a/livekit-rtc/livekit/rtc/room.py
+++ b/livekit-rtc/livekit/rtc/room.py
@@ -487,7 +487,9 @@ class Room(EventEmitter[EventTypes]):
 
         self._ffi_handle = FfiHandle(cb.connect.result.room.handle.id)
 
-        self._e2ee_manager = E2EEManager(self._ffi_handle.handle, options.e2ee)
+        self._e2ee_manager = E2EEManager(
+            self._ffi_handle.handle, options.encryption or options.e2ee
+        )
 
         self._info = cb.connect.result.room.info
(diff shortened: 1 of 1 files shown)


livekit-rtc/livekit/rtc/room.py
```diff
@@ -735,6 +762,14 @@ def _on_room_event(self, event: proto_room.RoomEvent):
                 changed_attributes,
                 participant,
             )
+        elif which == "participant_encryption_status_changed":
+            identity = event.participant_encryption_status_changed.participant_identity
+            participant = self._retrieve_participant(identity)
+            self.emit(
+                "participant_encryption_status_changed",
+                event.participant_encryption_status_changed.is_encrypted,
+                event.connection_quality_changed.quality,
```
xianshijing-lk: why we emit event.connection_quality_changed.quality here ? 

if that is needed, please put in a comment.

lukasIO: thanks for catching that, this should be emitting the participant at this position. Updated in https://github.com/livekit/python-sdks/pull/512/commits/2a78904404e518ff55a311b9fe132df3a2282aea

theomonnom:
Do we still use https://github.com/livekit/python-sdks/blob/main/livekit-rtc/livekit/rtc/e2ee.py ?

lukasIO:
Would have to take a closer look whether we can safely remove it, but I think this can be done as a follow up

PR #533, opened 2025-12-02
Add frame processor support for audio streams

lukasIO:


livekit-rtc/livekit/rtc/frame_processor.py
```diff
@@ -0,0 +1,34 @@
+from abc import ABC, abstractmethod
+from typing import Generic, TypeVar, Union
+from .audio_frame import AudioFrame
+from .video_frame import VideoFrame
+
+
+T = TypeVar("T", bound=Union[AudioFrame, VideoFrame])
+
+
+class FrameProcessor(Generic[T], ABC):
```
ladvoc, line 10: suggestion (non-blocking): This might be over abstracting, but we could make this more generalizable by defining an _AuthenticatedFrameProcessor_ interface which inherits from a more general _FrameProcessor_ one.

lukasIO: true, we can always go the other way around later with an `Unauthenticated...` (not that name, but something better) that stubs out the setCredentials call.

xianshijing-lk (approved):

livekit-rtc/livekit/rtc/audio_stream.py
```diff
@@ -62,7 +63,7 @@ def __init__(
         sample_rate: int = 48000,
         num_channels: int = 1,
         frame_size_ms: int | None = None,
-        noise_cancellation: Optional[NoiseCancellationOptions] = None,
+        noise_cancellation: Optional[NoiseCancellationOptions | FrameProcessor[AudioFrame]] = None,
```
theomonnom, line 67: Should we have another field name than `noise_cancellation`?
Like just `processors`?

livekit-rtc/livekit/rtc/frame_processor.py
```diff
@@ -0,0 +1,34 @@
...
+    ): ...
+
+    @abstractmethod
+    def _update_credentials(self, *, token: str, url: str): ...
+
+    @abstractmethod
+    def _process(self, frame: T) -> T: ...
+
+    @abstractmethod
+    def _close(self): ...
```
theomonnom: Do we need those methods (e.g: `_update_credentials`) ? Seems like an implementation detail?

theomonnom: I'm thinking this interface could be used by anybody, more like a general purpose API

livekit-rtc/livekit/rtc/audio_stream.py
```diff
@@ -268,6 +272,8 @@ async def _run(self):
             if audio_event.HasField("frame_received"):
                 owned_buffer_info = audio_event.frame_received.frame
                 frame = AudioFrame._from_owned_info(owned_buffer_info)
+                if self._processor is not None:
+                    frame = self._processor._process(frame)
```
davidzhao: neat.. this is clean

livekit-rtc/livekit/rtc/frame_processor.py
davidzhao: I think the underlying Rust implementation would need to make calls to `_update_credentials` when tokens have refreshed.. this is the same thread as [jacob's comment](https://github.com/livekit/python-sdks/pull/533#discussion_r2581332970)

livekit-rtc/livekit/rtc/audio_stream.py
lukasIO: The reason I went with reusing this for now is that the first official processor we'll have will also be noise cancellation, so I thought it might be confusing otherwise. 

I think we can still transition away (deprecate) noise_cancellation lateron in favour of something more generic

livekit-rtc/livekit/rtc/frame_processor.py
lukasIO: > I'm thinking this interface could be used by anybody, more like a general purpose API

yeah, that would be good. 

The methods defined right now is the minimal API we need to make authenticated processors work without tying them too deeply to the RTC package.
Apart from that it's pretty much general purpose already.

livekit-rtc/livekit/rtc/frame_processor.py
```diff
@@ -0,0 +1,34 @@
...
+T = TypeVar("T", bound=Union[AudioFrame, VideoFrame])
+
+
+class FrameProcessor(Generic[T], ABC):
+    @property
+    @abstractmethod
+    def is_enabled(self) -> bool: ...
+
+    @abstractmethod
+    def set_enabled(self, enable: bool): ...
```
theomonnom: Do you know if using a property would work? It seems more python idiomatic

```
    @property
    @abstractmethod
    def enabled(self) -> bool: ...

    @enabled.setter
    @abstractmethod
    def enabled(self, value: bool) -> None: ...
```

livekit-rtc/livekit/rtc/frame_processor.py
```diff
@@ -0,0 +1,34 @@
...
+    def _update_stream_info(
+        self,
+        *,
+        room_name: str,
+        participant_identity: str,
+        publication_sid: str,
+    ): ...
+
+    @abstractmethod
+    def _update_credentials(self, *, token: str, url: str): ...
```
theomonnom: nit: If this is called by an external event, maybe we can call those events like this? 
So it is obvious for the users implementing them that this is external. 
```
def on_stream_info_updated()

def on_credentials_updated()
```

livekit-rtc/livekit/rtc/frame_processor.py
theomonnom, line 33: I think it's OK for close to be public
```suggestion
    def close(self): ...
```

livekit-rtc/livekit/rtc/frame_processor.py
theomonnom: Can be used like: 

processor.enabled = False
processor.enabled = True

livekit-rtc/livekit/rtc/frame_processor.py
theomonnom: Let's just have a default no-op impl for this (removed `@abstractmethod`) 
```suggestion
    def _update_stream_info(
        self,
        *,
        room_name: str,
        participant_identity: str,
        publication_sid: str,
    ): ...

    def _update_credentials(self, *, token: str, url: str): ...
```

livekit-rtc/livekit/rtc/frame_processor.py
lukasIO: I was thinking `close` would rather be an internal method that we call to end the lifecycle of a processor.

Users would be expected to disable it with  `processor.enabled = False`.

livekit-rtc/livekit/rtc/frame_processor.py
lukasIO: sounds good

livekit-rtc/livekit/rtc/frame_processor.py
lukasIO: sounds fine to me!

livekit-rtc/livekit/rtc/frame_processor.py
lukasIO: applied @theomonnom's suggestion to have a default empty implementation for these

livekit-rtc/livekit/rtc/frame_processor.py
lukasIO: this is handled now by having default empty implementations for the credentials and stream info updates

livekit-rtc/pyproject.toml
```diff
@@ -20,3 +20,6 @@ manylinux-s390x-image = "manylinux_2_28"
 manylinux-pypy_x86_64-image = "manylinux_2_28"
 manylinux-pypy_i686-image = "manylinux_2_28"
 manylinux-pypy_aarch64-image = "manylinux_2_28"
+
+[tool.cibuildwheel.linux]
+before-build = "pip install requests && python rust-sdks/download_ffi.py --output livekit/rtc/resources && yum install -y libX11-devel libXrandr-devel libXext-devel libXfixes-devel libXdamage-devel libXcomposite-devel libXi-devel mesa-libGL-devel || apt install -y xorg-dev"
```
lukasIO, line 25: @theomonnom calling this out explicitly for you to take a look. 
After updating FFI we need these deps for the desktop capturer and the wheels repair command was failing with out them.

theomonnom: I think it's fine, but we should make sure the SDK is still working without users having to install those deps in their prod environment

PR #551, opened 2025-12-19
Strict type checking fixes

lukasIO:
with the uv setup working we got a bunch of type errors as a result of strict type checking, this PR enables strict type checking and fixes those type errors

### DIFF
--- a/livekit-rtc/livekit/rtc/media_devices.py
+++ b/livekit-rtc/livekit/rtc/media_devices.py
@@ -22,7 +22,7 @@ import numpy as np
 import threading
 
 if TYPE_CHECKING:
-    import sounddevice as sd  # type: ignore[import-not-found, import-untyped]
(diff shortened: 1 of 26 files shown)


livekit-rtc/livekit/rtc/media_devices.py
```diff
@@ -163,7 +163,7 @@ def __init__(
         output_device: Optional[int] = None,
         delay_estimator: Optional[_APMDelayEstimator] = None,
     ) -> None:
-        import sounddevice as sd  # type: ignore[import-not-found, import-untyped]
+        import sounddevice as sd
```
lukasIO, line 166: @theomonnom was unsure if we still need the ignore directives for sounddevice?

xianshijing-lk (approved):

theomonnom:
uh i may have closed too early, but this PR seems stale

PR #646, opened 2026-04-29
E2E Test: select video quality when publish simulcast video track.

cloudwebrtc:
## Summary

Adds an end-to-end pytest that verifies LiveKit simulcast layer switching from a Python publisher across both VP8 and H.264 codecs.

The test publishes a 1280×720 video track (rolling colored bars, generated as I420 with numpy) with `simulcast=True`, subscribes from a second room, and then drives the receiver through `RemoteTrackPublication.set_video_quality(HIGH/MEDIUM/LOW)`. For each layer it asserts the decoded frames converge to roughly the expected dimensions (within ±20% tolerance, requiring 5 consecutive matching frames):

| Quality   | Layer | Expected     |
|-----------|-------|--------------|
| `HIGH`    | `f`   | 1280×720     |
| `MEDIUM`  | `h`   | 640×360      |
| `LOW`     | `q`   | 320×180      |

Parameterized over `VP8` and `H264`. A 5-second post-subscription warmup is included so the encoder and bandwidth estimator stabilize at the highest layer before quality switches begin.

## Dependency

This test depends on https://github.com/livekit/rust-sdks/pull/1052, which fixes `TrackPublication.simulcasted` to derive from the non-deprecated `TrackInfo.codecs[*].layers` field. Without that fix, modern LiveKit servers (1.10+) report `info.simulcast=false` (deprecated field), so `set_video_quality` becomes a silent no-op and this test fails on all non-`HIGH` layers.

**This E2E test will only pass after `livekit/rust-sdks#1052` is merged and a new `livekit-ffi` release that includes it is published and consumed by `livekit-rtc`.** Until then, the test should be expected to fail in CI (or be marked `xfail` / skipped at the workflow level).

## Requires

- `LIVEKIT_URL`, `LIVEKIT_API_KEY`, `LIVEKIT_API_SECRET` env vars (test is auto-skipped otherwise).
- `numpy` (already a dependency of `livekit-rtc`).
- `livekit-ffi` release containing https://github.com/livekit/rust-sdks/pull/1052.

## Test plan

- [ ] `pytest livekit-rtc/tests/test_change_video_quality.py -v -s` passes against a current LiveKit server (verified locally on server `1.10.1` with a patched FFI dylib, both VP8 and H264).
- [ ] Test is skipped cleanly when LiveKit env vars are missing.
- [ ] No regressions in the existing `tests/` suite.
- [ ] Re-run after `rust-sdks#1052` is released and the FFI binary is updated.


cloudwebrtc:
@copilot resolve the merge conflicts in this pull request

cloudwebrtc:
@copilot resolve the merge conflicts in this pull request

livekit-rtc/livekit/rtc/track_publication.py
```diff
@@ -112,5 +115,17 @@ def set_subscribed(self, subscribed: bool):
         req.set_subscribed.publication_handle = self._ffi_handle.handle
         FfiClient.instance.request(req)
 
+    def set_video_quality(self, quality: "VideoQuality.ValueType") -> None:
+        """For simulcasted video tracks, request a specific simulcast layer
+        from the server. Use one of `rtc.VideoQuality.VIDEO_QUALITY_LOW` (q),
+        `VIDEO_QUALITY_MEDIUM` (h), or `VIDEO_QUALITY_HIGH` (f).
+
+        This is a no-op (the SDK logs a warning) if the publication is not
+        simulcasted."""
```
xianshijing-lk: print out a warning if that is not simulcasted ?

livekit-rtc/tests/test_change_video_quality.py
```diff
@@ -0,0 +1,341 @@
+"""End-to-end Test for simulcast / SVC video quality layers.
```
xianshijing-lk, line 15: should you have copy right on this file ?

livekit-rtc/tests/test_change_video_quality.py
```diff
@@ -0,0 +1,341 @@
...
+        )
+        .to_jwt()
+    )
+
+
+def unique_room_name(base: str) -> str:
+    return f"{base}-{uuid.uuid4().hex[:8]}"
+
+
+async def _wait_until(
```
xianshijing-lk, line 103: no need to change the code now, but if this piece of code can be shared with other tests, we can move them into some utils

livekit-rtc/tests/test_change_video_quality.py
```diff
@@ -0,0 +1,341 @@
...
+    loop = asyncio.get_event_loop()
+    fut: asyncio.Future = loop.create_future()
+
+    def _on_event(*args, **kwargs) -> None:
+        if fut.done():
+            return
+        if predicate is None or predicate(*args, **kwargs):
+            fut.set_result(args)
+
+    room.on(event, _on_event)
```
xianshijing-lk, line 159: curiously, do you need to unregisters it after the future completes

livekit-rtc/tests/test_change_video_quality.py
```diff
@@ -0,0 +1,341 @@
...
+    )
+    return holder["pub"]
+
+
+def _expect_event(
+    room: rtc.Room,
+    event: EventTypes,
+    predicate: Optional[Callable[..., bool]] = None,
+) -> asyncio.Future:
+    loop = asyncio.get_event_loop()
```
xianshijing-lk: nit, consider get_running_loop() ? 

From quick AI search:
In Python's asyncio, get_running_loop() is the modern and preferred method for accessing an active event loop from within asynchronous code. get_event_loop() is a legacy function with complex behavior that has been deprecated for most use cases in recent Python versions.

livekit-rtc/tests/test_change_video_quality.py
```diff
@@ -0,0 +1,341 @@
...
+    return rtc.VideoFrame(width, height, rtc.VideoBufferType.I420, data.tobytes())
+
+
+async def _publish_loop(source: rtc.VideoSource, stop: asyncio.Event) -> None:
+    interval = 1.0 / PUBLISH_FPS
+    start = time.monotonic()
+    while not stop.is_set():
+        t = time.monotonic() - start
+        frame = _make_rolling_i420(PUBLISH_WIDTH, PUBLISH_HEIGHT, t)
+        source.capture_frame(frame)
```
xianshijing-lk, line 200: curiously, is capture_frame async or sync ? do you need await ?

livekit-rtc/tests/test_change_video_quality.py
```diff
@@ -0,0 +1,341 @@
...
+            f"{local_pub.width}x{local_pub.height}"
+        )
+        remote_pub = await _ensure_track_subscribed(receiver, local_pub.sid)
+        assert remote_pub.track is not None
+
+        # Give the SFU a moment to propagate layer metadata and let the
+        # encoder/bandwidth estimator ramp up to all layers before we start
+        # switching qualities. SVC codecs (VP9/AV1) typically need more
+        # time than VP8/H264 simulcast to produce all spatial layers.
+        await asyncio.sleep(10.0 if mode == "svc" else 5.0)
```
xianshijing-lk: this code smells a bit, it hard code to sleep some time ? any way to improve it ?

livekit-rtc/tests/test_change_video_quality.py
```diff
@@ -0,0 +1,341 @@
...
+            remote_pub.set_video_quality(quality)
+            ew, eh = LAYER_DIMENSIONS[layer]
+            actual = await _wait_for_layer(stream, ew, eh, timeout=20.0)
+            print(f"[{codec_name}] layer={layer} expected~{ew}x{eh} got={actual[0]}x{actual[1]}")
+    finally:
+        stop.set()
+        try:
+            await pub_task
+        except Exception:
+            pass
```
xianshijing-lk: question, why pass ? should we handle the Exception better?

livekit-rtc/tests/test_change_video_quality.py
cloudwebrtc: This is a sync method, so I think we don't need to await here.

livekit-rtc/tests/test_change_video_quality.py
cloudwebrtc: After testing and confirming again, I removed the `sleep` here; it seems it's not necessary.

livekit-rtc/tests/test_change_video_quality.py
cloudwebrtc: Yeah, later I'll add all the common functions to a single utility .py file after all E2E use cases are added.

cloudwebrtc:
@copilot resolve the merge conflicts in this pull request

xianshijing-lk (approved):

PR #647, opened 2026-04-29
E2E Test: data channel send/receive test.

cloudwebrtc:
## Summary

Adds end-to-end pytests for LiveKit's data-channel (`publish_data`) routing semantics, covering targeted delivery, broadcast, topic propagation, and topic-on-targeted sends. Each scenario uses dedicated `rtc.Room` instances per participant and asserts that only the intended receivers see the packet.

| Test | What it covers |
|---|---|
| `test_data_one_to_one` | `destination_identities=["receiver"]` reaches only `receiver`; bystander gets nothing. |
| `test_data_one_to_many_targeted` | `destination_identities=["r1", "r2"]` reaches both; an excluded participant gets nothing. |
| `test_data_broadcast` | Empty `destination_identities` fan-outs to every other participant in the room. |
| `test_data_topic_passthrough` | `topic` is preserved end-to-end; receiver-side filtering by topic works. |
| `test_data_targeted_with_topic` | Targeted send carries `topic`; non-targets receive nothing. |

Implementation notes:

- `_DataCollector` subscribes to `data_received` and (optionally) filters by sender identity, so cross-test packet noise is impossible.
- `_ensure_visible` waits until the sender's room state lists the target identities — without this, targeted publishes can race the participant-joined signal and silently drop.
- `_assert_no_data` uses a 1s settle window before asserting absence, to avoid false negatives from in-flight delivery.

## Requires

- `LIVEKIT_URL`, `LIVEKIT_API_KEY`, `LIVEKIT_API_SECRET` env vars (tests are auto-skipped otherwise).

## Test plan

- [ ] `pytest livekit-rtc/tests/test_dc.py -v` passes against a current LiveKit server (verified locally on `1.10.1`).
- [ ] Tests are skipped cleanly when LiveKit env vars are missing.
- [ ] No regressions in the existing `tests/` suite.


livekit-rtc/tests/test_dc.py
```diff
@@ -0,0 +1,373 @@
...
+
+def skip_if_no_credentials():
+    required_vars = ["LIVEKIT_URL", "LIVEKIT_API_KEY", "LIVEKIT_API_SECRET"]
+    missing = [var for var in required_vars if not os.getenv(var)]
+    return pytest.mark.skipif(
+        bool(missing), reason=f"Missing environment variables: {', '.join(missing)}"
+    )
+
+
+def create_token(identity: str, room_name: str) -> str:
```
xianshijing-lk: I have been seeing many of these create_token() / skip_if_no_credentials() functions in different tests.

Should we come up with a tests/utils.py that consolidate all the util code in one place rather than duplicating them ?

cloudwebrtc: Created related PR here.
https://github.com/livekit/python-sdks/pull/684

lukasIO (approved):

PR #679, opened 2026-05-20
Fully configure frame processors when they are used directly on an audio stream

1egoman:
Updates the python sdk so that `FrameProcessor`-based noise cancellation providers can be used directly on `AudioStream`, without having to go through the agent's RoomIO to be able to initialize itself with credentials.

For example, with this change, something like the below becomes possible:
```python
stream = rtc.AudioStream.from_track(                                                                                                                   
    track=track,
    sample_rate=SAMPLE_RATE,                                             
    num_channels=CHANNELS,
    noise_cancellation=ai_coustics.audio_enhancement(model=ai_coustics.EnhancerModel.QUAIL_VF_L)  ,
) 
```

The way this works - `Track`s now keep track of which room they are part of (holding a `weakref` value). When the room a track is in changes, it computes new frame processor options and sends these to any `AudioStream`s which are associated with the track.

The `noise_cancellation_leave_open` parameter allows the agents sdk to call this `from_track` method with a frame processor which remains open across the whole session, and won't be auto-closed when the track is closed.

This goes along with https://github.com/livekit/agents/pull/5867, which removes the relevant event handling logic in the agents sdk. I will follow up with a node version of this once the python one is in a good state.

## Todo
- [x] Add some tests for this newly added behavior


livekit-rtc/livekit/rtc/track.py
```diff
@@ -12,21 +12,98 @@
...
+            # track left a room — clear processor's room context
+            # FIXME: This isn't really good, and I can't figure out what should happen here
+            # Closing the processor doesn't work (the track could get added to another room later)
+            # Empty values like this don't work, because it causes a drm::Error in the plugin
+            # Talk to lukas about this in a 1:1 and see if he can think of anything better
+            stream._on_processor_stream_info_updated(
+                room_name="", participant_identity="", publication_sid=""
+            )
+            # stream._on_processor_credentials_updated(token="", url="")
+            return
```
1egoman: This is the biggest thing still to be determined: I'm not exactly sure what to do when a track is removed from a room while an `AudioStream` that is within that track has a `FrameProcessor` registered.

- Initial thought:  set empty metadata with something like `stream._on_processor_credentials_updated(token="", url="")`. However, this causes the ai-coustics plugin to throw an error because the `""` string cannot be parsed as a URL.
- Next thought: disable the FrameProcessor by setting `enabled` to `False` in this situation (probably also log a warning too that this is being done?). The big problem with this is it would overzelously disable `FrameProcessor`s which don't need credentials to work (like the already-existing Krisp VIVA `FrameProcessor`).
- Other idea: Maybe modify the `FrameProcessor` interface to make `token` / `url` `Optional[str]`? But that would be a breaking api change, so that's probably out...

Curious what others think here and if there's an approach which I have missed.

lukasIO: I'm curious about this part 

> Closing the processor doesn't work (the track could get added to another room later)

what prevents us from re-instantiating the processor after it's been closed when/if it gets published to a new room?

Tangentially related (and sorry for opening that can of worms): how does this work if the same track gets published to multiple rooms in parallel ?

1egoman: Talked about this in a 1:1, a quick summary - I'm going to add a new `_on_credentials_cleared` event which will be backwards compatible and will allow this to be signaled so a plugin can clear its active credentials. Re-instantiating a `FrameProcessor` won't work because a class instance is being passed, not a class itself. I have done this in 07fec79.

For this:
> how does this work if the same track gets published to multiple rooms in parallel ?

We discussed it briefly but this already a problem today if a single frame processor were used in two agents sdk instances. The behavior today which this would replicate would be "last one wins"; the credentials used would be the last set which were emitted by any room. I'm not sure if this is necessarily catastrophically bad though for admittedly a fairly significant edge case, and fixing this outright would require a lot more evolution to `FrameProcessor`. So I think I am going to leave it as is for now.

livekit-rtc/livekit/rtc/audio_stream.py
```diff
@@ -303,6 +339,10 @@ async def aclose(self) -> None:
         This method cleans up resources associated with the audio stream and waits for
         any pending operations to complete.
         """
+        if self._processor is not None and not self._processor_leave_open:
+            self._processor._close()
+        if self._track is not None:
+            self._track._unregister_audio_stream(self)
         self._ffi_handle.dispose()
         await self._task
```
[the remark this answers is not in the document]
1egoman: Fixed in 8d3f4fe

livekit-rtc/livekit/rtc/audio_stream.py
```diff
@@ -65,6 +65,7 @@ def __init__(
         num_channels: int = 1,
         frame_size_ms: int | None = None,
         noise_cancellation: Optional[NoiseCancellationOptions | FrameProcessor[AudioFrame]] = None,
+        noise_cancellation_leave_open: bool = False,
```
theomonnom:
```suggestion
```

Can we move that inside NoiseCancellationOptions?

1egoman: Unfortunately, no - this is important to the `FrameProcessor[AudioFrame]` side of that `noise_cancellation` union. Open to putting it somewhere else but it needs to be settable in the `FrameProcessor` path.

lukasIO: hmm, not sure if it's a good idea, but could it be a field on the `FrameProcessor` interface instead? 

Then we could add it to NoiseCancellationOptions and new FrameProcessors would be able to set it on the processor itself

1egoman: It's not a setting that a frame processor would always want to have set or not have set, so I'm not sure that would really make sense either.

For context, the reason this is here is so the agents sdk can reuse a single `FrameProcessor` across multiple underlying tracks. Previously, this wasn't a problem in the way this used to work, because the agents sdk had the responsibility of closing the `FrameProcessor`, so it could easily do it at room disconnection time. But in order to support the ability to use `FrameProcessor`s directly on an `AudioStream`, calling `close` needs to be pushed down deeper than the agents sdk layer. This flag allows the caller to explictly tell `AudioStream` that they will manage cleaning up the `FrameProcessor` so that both use cases can continue to work.

xianshijing-lk: I think this flag is not really configuring the noise suppression behavior, but how AudioStream deals with its own noise suppression, maybe the naming of noise_cancellation_leave_open  is a bit confusing ? 

how about close_noise_cancellation_on_stream_close or manage_noise_cancellation_processor ?

lukasIO: > It's not a setting that a frame processor would always want to have set or not have set

it could stay `undefined` by default? 🤷
I understand however that it feels a bit weird for it to live on the processor if the processor itself doesn't really use the field.

We shortly discussed also the option to introduce a `restart` method on the processor. I think this could still be a viable alternative?

1egoman: > We shortly discussed also the option to introduce a restart method on the processor. I think this could still be a viable alternative?

It could, but the con there is it's a breaking api change to `FrameProcessor`.

Just generally, I want to understand what folks' concerns are in more detail. Is it just the `noise_cancellation_` prefix naming like shijing suggested (I think out of the two suggestions, I like `manage_noise_cancellation_processor` better)? Or is there something deeper behavior wise that is concerning?

FWIW, two fairly similar patterns I found:
- `LiveKitAPI` conditionally controls `aiohttp.ClientSession` cleanup [here](https://github.com/livekit/python-sdks/blob/fbbe6f3a9941749aa19f2702cd4c5f66026bc154/livekit-api/livekit/api/livekit_api.py#L12) based on whether the user passes a custom session or uses an inbuilt session.
- The `LocalAudioTrack` has a `userProvidedTrack` parameter which is used to control whether the track is cleaned up or not [here](https://github.com/livekit/client-sdk-js/blob/55a3e4f708fa6b1152a41c46f0a36da2590956d5/src/room/track/LocalAudioTrack.ts#L39).

1egoman: Talked to @lukasIO in a 1:1 and he confirmed his concern was mostly with the naming, not with the broad approach, which is helpful.

A few other name ideas, in addition to shijing's suggestions (`close_noise_cancellation_on_stream_close` / `manage_noise_cancellation_processor`) - some of these would involve flipping the flag:
- `shared_noise_cancellation`
- `noise_cancellation_externally_managed`
- `auto_close_noise_cancellation`
- `owns_noise_cancellation`

Out of the above, I think I like `auto_close_noise_cancellation` the best:
```python
# Usage within agents sdk:
AudioStream.from_track(
    # ...
    noise_cancellation=frame_processor,
    auto_close_noise_cancellation=False,
)
```

I'm going to update the pull request to use it for now in 8d5e656.

----------

Another possible idea: maybe something like the below could be a different way to package the same data which could better contain it. In a world like this, `noise_cancellation` would be of type `Union[NoiseCancellationOptions, FrameProcessorOptions, FrameProcessor]`:
```python
AudioStream.from_track(
    # ...
    noise_cancellation=FrameProcessorOptions(frame_processor=self, leave_open=True)
)
```

Do any of these ideas look better than the current state?

1egoman: @theomonnom @xianshijing-lk Lukas gave a a general 👍 to the rename to `auto_close_noise_cancellation` addressing his concern. Do either of you have further concerns beyond what this rename could accomplish with this approach?

If not / I don't hear anything in the next few days I think I am good to merge this.

lukasIO (approved):

livekit-rtc/livekit/rtc/audio_stream.py
```diff
@@ -65,6 +65,7 @@ def __init__(
         num_channels: int = 1,
         frame_size_ms: int | None = None,
         noise_cancellation: Optional[NoiseCancellationOptions | FrameProcessor[AudioFrame]] = None,
+        auto_close_noise_cancellation: bool = True,
```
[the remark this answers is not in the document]
1egoman: Fixed in ea1de09

livekit-rtc/livekit/rtc/room.py
```diff
@@ -757,6 +759,8 @@ def _on_room_event(self, event: proto_room.RoomEvent) -> None:
             unpublished = self.local_participant._track_publications.get(sid)
             if unpublished is not None:
                 del self.local_participant._track_publications[sid]
+                if unpublished.track is not None:
+                    unpublished.track._set_room(None)
```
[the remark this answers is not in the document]
1egoman: Fixed in b9a9c15

livekit-rtc/livekit/rtc/room.py
```diff
@@ -778,6 +778,15 @@ def _on_room_event(self, event: proto_room.RoomEvent) -> None:
...
                 self.local_participant._track_publications[republished.sid] = republished
+                if republished.track is not None:
+                    # Keep the local-track invariant (track.sid == publication.sid,
+                    # set at publish_track) intact across republish, then re-push
+                    # metadata so any attached FrameProcessor learns the new
+                    # publication SID / credentials. _set_room with the same room
+                    # is a no-op for the token_refreshed listener but re-fans the
+                    # metadata to every registered AudioStream.
+                    republished.track._info.sid = republished.sid
+                    republished.track._set_room(self)
```
1egoman, line 801: Note to any reviewers: the Devin bot surfaced [this case](https://github.com/livekit/python-sdks/pull/679#discussion_r3374841676) where when a full reconnect happens, tracks would not have their metadata kept up to date. I'm not sure if `republished.track._info.sid = republished.sid` is a good pattern, but wanted to give an opportunity for folks to weigh in on these final late breaking changes before merging.

livekit-rtc/livekit/rtc/room.py
```diff
@@ -757,6 +759,14 @@ def _on_room_event(self, event: proto_room.RoomEvent) -> None:
...
                 del self.local_participant._track_publications[sid]
+                if unpublished.track is not None:
+                    unpublished.track._set_room(None)
+                    # Mirror track_unsubscribed: drop the publication's track
+                    # reference. This also makes unpublish_track's own
+                    # _set_room(None) a no-op when it loses the race (its
+                    # `publication._track is not None` guard short-circuits),
+                    # avoiding a redundant clear.
+                    unpublished._track = None
                 self.emit("local_track_unpublished", unpublished)
```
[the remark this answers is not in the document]
1egoman: This is sort of the [same thing that was surfaced here](https://github.com/livekit/python-sdks/pull/679#discussion_r3374841676) just flipped around?

I don't really have the depth of knowledge on this code to know for sure what to do here. Here's what a LLM had to say about this whole situation:
```
Is it actually a breaking change? Not meaningfully.

  Three reasons:

  1. It was never reliable before. Because of the documented unpublish race, unpublish_track could null publication._track before the room event handler
  ran. So pre-PR, a callback reading publication.track already saw None sometimes (race-dependent). Any correct consumer already had to handle None. We've
  gone from "racy/sometimes-set" to "always None" — not from "guaranteed-set" to "None."
  2. No consumer reads it. I grepped both repos for local_track_unpublished handlers:
    - python-sdks: tests/rtc/test_e2e.py and examples/basic_room.py — both read only publication.sid.
    - agents SDK: doesn't subscribe to local_track_unpublished at all.        

  So there's zero actual breakage today.                                      
  3. The event never offered guaranteed track access anyway. Unlike track_unsubscribed (which passes the track as a separate emit arg),
  local_track_unpublished only emits the publication. So publication.track was the only route, and it was always racy.

  Net: it's an observable behavior change to a public event's payload, but not a break that affects any real consumer, and it removes only an
  already-unreliable capability. I'd be comfortable merging as-is (optionally noting it in a changelog).

  If you want to be conservative, two options                                 

  - Reorder: null after the emit (one line). _on_room_event is synchronous and emit is synchronous, so the null still completes before any other coroutine
  (including unpublish_track) can interleave — meaning the redundant-clear avoidance still holds, and the callback can read publication.track in the
  non-racy case (restoring the exact pre-PR behavior). This is the minimal way to make Devin's concern moot without changing the event signature.
  - Add the track as a second emit arg (like track_unsubscribed): gives reliable access, but it's a public event-signature change — bigger, and I would not
  fold that into this PR.                                                     

  My recommendation: leave it as-is (nulling before emit is the more honest semantic — "the track is gone"), since nothing consumes it. But if you'd rather
  not introduce any behavior delta at all, the after-emit reorder is a safe one-liner
  ```

I'm tempted to leave it alone since this very minor "breaking change" I think is actually now in practice more correct. But I'm open to alternate perspectives here.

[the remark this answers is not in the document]
Comment (lukasIO) [reply livekit-rtc/livekit/rtc/room.py]: wouldn't the alternative be to simply null it after emitting the event? for sync event handlers that should keep the previous behaviour?

File: livekit-rtc/livekit/rtc/room.py (hunk 6 above)
Comment (lukasIO) [reply livekit-rtc/livekit/rtc/room.py:801]: didn't dive too deep into this, just a thought: if it's about the metadata being outdated, shouldn't we override _all_ of the track's `info` ?

Comment (1egoman) [reply livekit-rtc/livekit/rtc/room.py:801]: 🤔 Yea good point

File: livekit-rtc/livekit/rtc/room.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (1egoman) [reply livekit-rtc/livekit/rtc/room.py]: Yep, the LLM recommended that as an option and after thinking about it further I think this is probably the more conservative choice, done in adc16d3.

File: livekit-rtc/livekit/rtc/room.py (hunk 6 above)
Comment (1egoman) [reply livekit-rtc/livekit/rtc/room.py:801]: Dug into this a big more - actually, it's not quite that straightforward, since `republished.track._info` is a `TrackInfo`, but `local_track_republished.info` is a `TrackPublicationInfo`. The overlapping fields are fairly minimal and I think not things which would change on a full reconnect - `sid`, `name`, `kind`, `muted`, and `remote`. So I think this is good as is.

Comment (lukasIO) [reply livekit-rtc/livekit/rtc/room.py:801]: makes sense

File: livekit-rtc/livekit/rtc/frame_processor.py [hunk 8]
```diff
@@ -24,8 +24,12 @@ def _on_stream_info_updated(
         publication_sid: str,
     ) -> None: ...
 
+    def _on_stream_info_cleared(self) -> None: ...
+
     def _on_credentials_updated(self, *, token: str, url: str) -> None: ...
 
+    def _on_credentials_cleared(self) -> None: ...
```
lukasIO, line 31: question(non-blocking): do you foresee any scenario where `on_credentials_cleared` and `on_stream_info_cleared` would get called independently from one another? 


I don't have a good suggestion for a unified method name, just raising this in case you can think of one that would be obvious and allow for less methods to be implemented for each processor

lukasIO (approved):

livekit-rtc/livekit/rtc/frame_processor.py
1egoman: Yes, one example (at least in theory) where they could be called at different times: a frame processor is detached from a track, kept around connected to the room for a period of time, then reattached to another track in that same room.

1egoman:
Did a little bit of testing locally along with https://github.com/livekit/agents/pull/5867 and this looks to work in all the easy to test happy path cases. That along with the test coverage of the harder to test non happy path cases I think makes me fairly confident in this now!

Also, in case anyone is curious - I also have tested the current `main` of the agents repo (`451bae4f7`) with this loaded and everything works, there just are in some cases callbacks like `_on_credentials_updated ` / etc are being called twice (once by agents code, once by the code in this PR) instead of once.

PR #685, opened 2026-05-22
Add rpc v2 long info case to example

1egoman:
Ports the rpc v2 example case added in [node-sdks](https://github.com/livekit/node-sdks/blob/sderosa/BOT-344-ready-for-room-event/examples/rpc/index.ts) and [client-sdk-js](https://github.com/livekit/client-sdk-js/blob/main/examples/rpc/rpc-demo.ts) to this sdk as well.

### DIFF
--- a/examples/rpc.py
+++ b/examples/rpc.py
@@ -35,6 +35,12 @@ async def main():
             print("Error:", error)
 
         try:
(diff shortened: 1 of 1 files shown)


lukasIO (approved):
lgtm

PR #699, opened 2026-06-03
fix: unsubscribe FFI queue when the room listen task ends

longcw:
When a room is disconnected remotely (e.g. the participant is removed), `Room.disconnect()` early-returns on `not isconnected()` and never unsubscribes the room's FFI event queue. The FFI receive thread keeps posting events to it, and once the event loop closes it logs `error putting to queue: Event loop is closed`.

Unsubscribe the FFI queue in a done callback on the listen task, so the subscription's lifetime matches its only consumer (`_listen_task`) regardless of how it ends — EOS, error, or cancellation. `unsubscribe` is idempotent, so the existing calls in `disconnect()` and `__del__` stay harmless.

### DIFF
--- a/livekit-rtc/livekit/rtc/room.py
+++ b/livekit-rtc/livekit/rtc/room.py
@@ -557,6 +557,14 @@ class Room(EventEmitter[EventTypes]):
         # start listening to room events
         self._task = self._loop.create_task(self._listen_task())
 
+        # unsubscribe from the FFI queue once the listen task ends.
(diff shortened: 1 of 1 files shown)


theomonnom (approved):

cloudwebrtc (approved):

xianshijing-lk:
sorry for being late on review, it looks good to me too. Thanks for fixing it.

PR #713, opened 2026-06-12
Fixes a race where `await room.sid` could hang even after the room SID was available in `room._info.sid`.

xianshijing-lk:
The Python binding previously resolved `_first_sid_future` only from `room_sid_changed`. Under some event ordering, the SID could arrive through the initial connect result or `room_updated` first, leaving existing `room.sid` waiters
  unresolved.

  ## Changes

  - Shield `_first_sid_future` so caller-side `asyncio.wait_for(room.sid, ...)` timeouts do not cancel shared Room state.
  - Add `_resolve_first_sid()` helper.
  - Resolve the first SID future from:
    - initial connect result
    - `room_sid_changed`
    - `room_updated`
    - `moved`

  ## Testing

Run Theo's sid_repro.py

### DIFF
--- a/livekit-rtc/livekit/rtc/room.py
+++ b/livekit-rtc/livekit/rtc/room.py
@@ -202,7 +202,16 @@ class Room(EventEmitter[EventTypes]):
         if self._info.sid:
             return self._info.sid
 
-        return await self._first_sid_future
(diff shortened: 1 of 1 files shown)


cloudwebrtc (approved):

theomonnom:
Thanks @xianshijing-lk !

PR #725, opened 2026-06-22
feat(memory): add livekit-memory — sub-10ms in-process semantic memory

Piyussh01:
## Problem

Voice agents need a user's context *mid-turn*, under a tight latency budget. A remote vector DB round-trip or a transformer query-embed on CPU (~10ms median, ~50ms p99) is enough to break a live conversation loop.

## Approach

A new self-contained workspace package `livekit-memory` (`livekit.memory`) providing in-process semantic memory:

- **Static embedder** (`Model2VecEmbedder`) — token-lookup + mean-pool, no transformer forward pass. ~0.03ms per short query.
- **Swappable index** — exact `BruteForceIndex` (one normalized matmul, sub-ms to ~100k vectors), auto-upgrading to `UsearchIndex` (HNSW) past ~100k.
- **Facts vs. collection split** — a small always-scanned `facts` namespace (pinned user profile) plus an ANN-indexed semantic collection, with a `context()` helper that returns one prompt-ready string per turn.

## Evidence (measured, Apple M4 Pro, 384d)

| Stage | N=1M |
|---|---|
| Model2Vec embed | 0.028ms median |
| usearch HNSW search (~0.95 recall) | 0.27ms median / 0.54ms p99 |
| **End-to-end (embed + search)** | **0.17ms median / 0.31ms p99** |

~30x under the 10ms budget at a million vectors. Transformer embedders (fastembed MiniLM ~9.4ms median / ~53ms p99) confirm why the static embedder is required.

## Scope & safety

- New isolated package; **no changes** to `rtc` / `api` / `protocol`.
- Pure-Python core, numpy-only required dep. `model2vec` / `usearch` / `fastembed` are optional extras, env-gated to Python >= 3.10.
- Dependency-free `HashingEmbedder` + brute-force is the default, so tests run offline with no model downloads.
- Passes `make check` (ruff format, ruff lint, mypy --strict) and adds 11 tests.

## Caveats

- Static-embedding *retrieval* quality is ~82% of MiniLM's; the `onnx` extra (`CallableEmbedder` + fastembed) is the opt-in escape hatch for higher recall at higher latency.
- HNSW recall degrades under heavy churn — long-lived indices want periodic compaction.

Happy to discuss whether this belongs here vs. `livekit/agents`, and to trim to the pure-numpy core if preferred.


xianshijing-lk:
+ @longcw, could you please help take a look ? 

cc @theomonnom

Piyussh01:
following up on this @xianshijing-lk 

i think it'll be a nice addition to livekits product roadmap. we can try moving away from static embedding in the future

PR #733, opened 2026-06-27
feat: auto failover APIs with LK Cloud

davidzhao:
retries in alternative datacenters on 5xx and transport failures

### DIFF
--- a/livekit-api/livekit/api/twirp_client.py
+++ b/livekit-api/livekit/api/twirp_client.py
@@ -12,13 +12,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Dict, Type, TypeVar, Optional
+import asyncio
+from typing import Dict, List, Optional, Type, TypeVar
 
 import aiohttp
 from google.protobuf.message import Message
 from urllib.parse import urlparse
 
+from ._failover import (
+    FailoverConfig,
+    RegionCache,
+    failover_attempts,
+    failover_backoff_base,
+    host_key,
+    origin_of,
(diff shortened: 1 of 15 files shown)


theomonnom (approved):

livekit-api/livekit/api/twirp_client.py
```diff
@@ -92,6 +106,7 @@ def __init__(
         host: str,
         pkg: str,
         prefix: str = DEFAULT_PREFIX,
+        failover: Optional[FailoverConfig] = None,
```
theomonnom: nit: FailoverOptions for consistency with agents

theomonnom:
Should we fail fast with a shorter timeout?

livekit-api/livekit/api/twirp_client.py
```diff
@@ -114,21 +131,73 @@ async def request(
...
+                    if resp.status == 200:
+                        return response_class.FromString(await resp.read())
+                    # Twirp encodes errors as JSON regardless of content type.
+                    try:
+                        error_data = await resp.json()
+                    except Exception:
+                        error_data = {}
+                    if resp.status < 500:
+                        # 4xx is terminal.
+                        raise self._twirp_error(error_data, resp.status)
```
theomonnom, line 179: Should we log a warning on retry?

livekit-api/livekit/api/twirp_client.py
davidzhao: removed this entirely, should not be customizable other than to disable

PR #740, opened 2026-07-01
rtc: fail fast instead of hanging when FFI is used across fork()

theomonnom:
The native runtime (tokio worker/IO threads in liblivekit_ffi) does not survive `fork()`, and re-initializing only resets the event callback rather than rebuilding the runtime. A child process that inherited an already-initialized `FfiClient` singleton therefore hangs on every FFI request.

Guard `FfiClient.instance` with the owning pid and raise a clear `RuntimeError` instead of hanging. This does not affect `spawn`/`forkserver` (each process initializes its own FFI); it only turns the `fork`-after-init footgun into an actionable error.

### DIFF
--- a/livekit-rtc/livekit/rtc/_ffi_client.py
+++ b/livekit-rtc/livekit/rtc/_ffi_client.py
@@ -83,7 +83,11 @@ class FfiHandle:
     def dispose(self) -> None:
         if self.handle != INVALID_HANDLE and not self._disposed:
             self._disposed = True
-            assert FfiClient.instance._ffi_lib.livekit_ffi_drop_handle(ctypes.c_uint64(self.handle))
+            ffi = FfiClient._instance
+            if ffi is None or ffi._pid != os.getpid():
+                return
+            dropped = ffi._ffi_lib.livekit_ffi_drop_handle(ctypes.c_uint64(self.handle))
+            assert dropped
 
     def __repr__(self) -> str:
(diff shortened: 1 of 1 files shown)


davidzhao (approved):
lgtm

livekit-rtc/livekit/rtc/_ffi_client.py
```diff
@@ -83,7 +83,9 @@ def disposed(self) -> bool:
     def dispose(self) -> None:
         if self.handle != INVALID_HANDLE and not self._disposed:
             self._disposed = True
-            assert FfiClient.instance._ffi_lib.livekit_ffi_drop_handle(ctypes.c_uint64(self.handle))
+            ffi = FfiClient._owned_instance()
+            if ffi is not None:
+                assert ffi._ffi_lib.livekit_ffi_drop_handle(ctypes.c_uint64(self.handle))
```
xianshijing-lk: will assert be stripped out if our customers use some optimization flags ? 

if that gets stripped out, I will assume ffi._ffi_lib.livekit_ffi_drop_handle will never be called.

xianshijing-lk (approved):
lgtm assuming you will address the comment about 'assert'

xianshijing-lk:
oh, one question, is there any way to add a test for the fork behavior ? so that we can prevent regression

davidmonnom (approved):

PR #743, opened 2026-07-03
avoid KeyError when track is already unpublished

davidzhao:
fixes #4639

PR #758, opened 2026-07-18
fix(rtc): guard remote track event handlers against racing disconnect

### LINKED ISSUE #757: KeyError in _on_room_event for track_subscribed when participant already removed by a racing disconnect
### Summary

`Room._on_room_event` raises `KeyError` in the `track_subscribed` handler when the remote participant has already been removed from `self._remote_participants` by a racing `participant_disconnected` (or a duplicate event). This is the same class of race that #743 fixed for `track_unpublished`, but the sibling handlers were left using an unguarded dict lookup.

### Traceback (observed in production)

```
KeyError: 'sip-caller-+91XXXXXXXXXX'
  File ".../livekit/rtc/room.py", in _listen_task
    self._on_room_event(event.room_event)
  File ".../livekit/rtc/room.py", in _on_room_event
    rparticipant = self._remote_participants[event.track_subscribed.participant_identity]
```

It is caught by the `except Exception: logging.exception(...)` in `_listen_task`, so it does not crash the session, but it is logged as an error and reported to Sentry as noise. It shows up most often with SIP participants, whose connect/disconnect ordering is racier than typical WebRTC clients (a fast caller hangup interleaves `participant_disconnected` with an in-flight `track_subscribed`).

### Root cause

On `main` (livekit-rtc 1.1.13), `track_unpublished` was made defensive in #743 with an explicit comment:

```python
# The participant or publication may already have been removed by a
# racing disconnect or a duplicate event, so both lookups are done
# defensively ... instead of raising a KeyError that _listen_task logs as an error.
rp = self._remote_participants.get(identity)
if rp is not None:
    ...
```

but the following handlers in the same method still use a raw `self._remote_participants[identity]` and hit the identical race:

- `track_published`
- `track_subscribed`  ← the one in the traceback above
- `track_unsubscribed`
- `track_subscription_failed`

### Expected behavior

These handlers should look the participant (and publication) up defensively and skip the emit when the entry is already gone, mirroring `track_unpublished` / `local_track_unpublished`, rather than raising a `KeyError`.

### Environment

- livekit-rtc: 1.1.8 (observed); confirmed still present on `main` (1.1.13)
- Python 3.11

Happy to send a PR mirroring the #743 approach.

AKomplished-bug:
Fixes #757

### Problem

`Room._on_room_event` raises `KeyError` in the `track_subscribed` handler (and its siblings) when the remote participant has already been removed from `self._remote_participants` by a racing `participant_disconnected` or a duplicate event:

```
KeyError: 'sip-caller-+91XXXXXXXXXX'
  File ".../livekit/rtc/room.py", in _on_room_event
    rparticipant = self._remote_participants[event.track_subscribed.participant_identity]
```

It is swallowed by the `except Exception: logging.exception(...)` in `_listen_task`, so it does not break the session, but it is logged as an error (and reported to error trackers as noise). It shows up most often with SIP participants, whose connect/disconnect ordering is racier than typical WebRTC clients.

### Fix

#743 already made `track_unpublished` defensive for exactly this race. This PR applies the same pattern to the remaining handlers that still used an unguarded `self._remote_participants[identity]`:

- `track_published`
- `track_subscribed`
- `track_unsubscribed`
- `track_subscription_failed`

Each now looks the participant (and, where applicable, the publication) up with `.get(...)` and skips the emit with a `logger.debug` when the entry is already gone, matching the wording and behaviour of the existing `track_unpublished` handler.

No behavioural change on the happy path; only the racing-disconnect case is affected.

### DIFF
--- a/livekit-rtc/livekit/rtc/room.py
+++ b/livekit-rtc/livekit/rtc/room.py
@@ -817,10 +817,19 @@ class Room(EventEmitter[EventTypes]):
             lpublication._first_subscription.set_result(None)
             self.emit("local_track_subscribed", lpublication.track)
         elif which == "track_published":
-            rparticipant = self._remote_participants[event.track_published.participant_identity]
-            rpublication = RemoteTrackPublication(event.track_published.publication)
-            rparticipant._track_publications[rpublication.sid] = rpublication
-            self.emit("track_published", rpublication, rparticipant)
+            # The participant may already have been removed by a racing
+            # disconnect or a duplicate event, so the lookup is done defensively
+            # and the emit is skipped when the participant is gone, mirroring the
+            # track_unpublished handler, instead of raising a KeyError that
+            # _listen_task logs as an error.
+            identity = event.track_published.participant_identity
+            published_participant = self._remote_participants.get(identity)
+            if published_participant is not None:
+                published = RemoteTrackPublication(event.track_published.publication)
+                published_participant._track_publications[published.sid] = published
+                self.emit("track_published", published, published_participant)
+            else:
+                logger.debug("track_published for untracked participant %s", identity)
         elif which == "track_unpublished":
             # The participant or publication may already have been removed by a
             # racing disconnect or a duplicate event, so both lookups are done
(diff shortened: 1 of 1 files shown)


AKomplished-bug:
Looks like there is some issue with the pipeline @davidzhao

xianshijing-lk (approved):
lgtm

xianshijing-lk:
I guess the CI problem is that it tries to land the PR to a fork (livekit:main) ? can you change it to land to main ?

AKomplished-bug:
@xianshijing-lk the base is already `livekit/python-sdks:main` — the failure is the `Generating protobuf` job's checkout using `ref` without `repository:`, so for fork PRs it looks for the branch in the base repo and the fetch 404s.

xianshijing-lk:
> @xianshijing-lk the base is already `livekit/python-sdks:main` — the failure is the `Generating protobuf` job's checkout using `ref` without `repository:`, so for fork PRs it looks for the branch in the base repo and the fetch 404s.

You are right, I am making a separate PR to fix it. Lets land this PR as is .

PR #766, opened 2026-07-23
Bump livekit ffi to 0.12.73

1egoman:
Updates `livekit-ffi` to the latest version, which contains data streams v2.

In addition to the ffi package update itself, there's a bit more that has to be done here because the python sdk today has its own data streams implementation. I'm working on [migrating this package to use the `livekit-ffi` provided data streams](https://github.com/livekit/python-sdks/pull/769) but in the meantime, if the `livekit-ffi` update was done today, then it would mean that the newly introduced `client_protocol` of `2` would be advertised, and the pre-existing python implementation would receive data stream v2 messages, not v1 (ie, single packet data streams, compression, etc could be sent to the existing python data streams code).

To address this, I've added a temporary livekit-ffi specific `options.data_stream.use_legacy_client_implementation` flag which can be set to disable advertising the new client protocol. What this means in practice:
- Outgoing data streams will all be v1 (uses python data streams implementation)
- Incoming data streams will all be v1 (other participants see `client_protocol` of `1`, so no inline content / compression)
- Outgoing RPCs will be potentially v1 or v2 (this within the rust code is powered by `livekit-ffi`'s data streams implementation, just like it works today)
- Incoming RPCs will also potentially be data streams v1 or data streams v2 powered (for the same reason)

Once `livekit-ffi` data stream migration is done in the above linked pull request, I'll merge it and drop this flag both here and on the rust sdk end, and then everything will use data streams v2 everywhere.

I've tested this locally (all the data streams and rpc paths above) with success.

## Todo
- [x] Merge rust data streams v2 pull request: https://github.com/livekit/rust-sdks/pull/1192
- [x] Do `livekit-ffi` release
- [x] Bump version in here, then ci should pass
- [x] Do final round of testing with everything pointed at the final release builds


.github/workflows/ruff.yml
```diff
@@ -18,8 +18,10 @@ jobs:
...
         run: uv sync --all-extras --dev
 
+      # Use the ruff version pinned in uv.lock rather than uvx's latest release,
+      # so lint results only change when the lockfile is intentionally updated.
       - name: Ruff livekit-api
-        run: uvx ruff check --output-format=github .
+        run: uv run ruff check --output-format=github .
 
       - name: Check format
-        run: uvx ruff format --check .
```
1egoman, line 25: This is a bit of a drive by fix, but I noticed that running `ruff` otherwise locally resulted in different formatted code then what I was getting in CI (which left the ci build in a failing state)

1egoman:
Did a final round of light testing with a data streams v2 capable client:
- Data streams sent manually from an agent running this build are v1
- RPC requests sent manually from an agent are v2

I did some ac hoc conversation back and forth and the whole agent lifecycle seemed to work fine. I could use some help validating anything more in depth though that uses data streams, like any specific plugins / etc.

lukasIO (approved):

theomonnom (approved):

PR #769, opened 2026-07-28
Port data streams over to rust livekit-ffi version

1egoman:
(The python version of https://github.com/livekit/node-sdks/pull/697)

Previously, the python sdk had its own data streams implementation. My understanding is there isn't really a good reason for this, and it's more of an artifact of history - the node/python sdks had data streams added first, and the rust implementation came later on.

With [data streams v2](https://github.com/livekit/client-sdk-js/pull/1985), the rust implementation will both be quite a bit more stable and gain some new features that make it significantly more performant (single packet data streams and DEFLATE compression when it makes the payload smaller). This roughly _doubles_ data stream throughput in local testing.

So, port the python sdk to use the rust sdk data streams v2 implementation, and completely remove the typescript implementation. This is being done via the already-existing `livekit-ffi` data streams messages which the C++ and unity sdks use today. This is a substantial change which needs thorough testing.

## New behaviors worth being aware of
All of these are either data streams v2 related changes, or bug fixes.

### 1. `compress` option
When sending a data stream, there is a new `compress` option. Just like how this works on web / rust, this defaults to `true`. If set to `false`, then compression will be disabled (useful if you know the data you are sending isn't compressible / you are doing your own compression, which is not uncommon in robotics use cases). The vast majority of users should leave this set to `true`.

### 2. Max data stream size
In a [rust data streams v2 pull request review comment](https://github.com/livekit/rust-sdks/pull/1192#discussion_r3571761970), we decided that for security reasons it made sense to introduce a maximum data stream size as a DOS protection. This limit is by default `5gb` - any data stream that is larger will now read up until that point, and if the stream keeps going, a "payload too large" error will be raised on the stream and exposed to a user on the subsequent `.read()` call.

If a user is sending a large file, they can override this by setting a new `maxPayloadByteLength` option on the `room.connect` call:
```python
room.connect(url, token, auto_subscribe=True, dynacast=False, data_stream=DataStreamOptions(max_payload_byte_length=1000))
```

### 3. Throwing data stream errors rather than exclusively logging them
The old python specific data streams implementation didn't properly surface errors to the caller which were encountered while reading the stream. As a prerequisite for exposing the error for item 2, I fixed this, but it means that now if a data stream were to fail, errors would get thrown when in the past they would get logged and dropped. I've also introduced a new `StreamError` as a new error type for these newly thrown errors.

## Testing
I've tested this as best as I can locally but I'd really like some help from agents folks to really put this through the ringer and make sure I haven't inadvertently broken something.

## Todo
- [x] Test this throroughly
- [x] Merge data streams v2 and do a rust release - https://github.com/livekit/client-sdk-js/pull/1985)
- [x] Bump the rust ffi version here and do another round of testing

### DIFF
--- a/livekit-rtc/livekit/rtc/data_stream.py
+++ b/livekit-rtc/livekit/rtc/data_stream.py
@@ -70,7 +70,7 @@ def _text_stream_info_from_proto(info: proto_data_stream.TextStreamInfo) -> Text
         mime_type=info.mime_type,
         topic=info.topic,
         timestamp=info.timestamp,
-        size=info.total_length if info.HasField("total_length") else None,
+        size=info.total_length if info.HasField("total_length") else 0,
         attributes=dict(info.attributes),
         attachments=list(info.attached_stream_ids),
     )
@@ -82,7 +82,7 @@ def _byte_stream_info_from_proto(info: proto_data_stream.ByteStreamInfo) -> Byte
         mime_type=info.mime_type,
         topic=info.topic,
         timestamp=info.timestamp,
-        size=info.total_length if info.HasField("total_length") else None,
+        size=info.total_length if info.HasField("total_length") else 0,
         attributes=dict(info.attributes),
         name=info.name,
     )
--- a/livekit-rtc/livekit/rtc/data_stream.py
+++ b/livekit-rtc/livekit/rtc/data_stream.py
@@ -89,6 +89,25 @@ def _byte_stream_info_from_proto(info: proto_data_stream.ByteStreamInfo) -> Byte
 
 
 class TextStreamReader:
+    """An incoming text stream.
+
+    Use as an async iterator to receive chunks, or :meth:`read_all` to collect
+    the whole payload::
+
+        async for chunk in reader:
+            process(chunk)
+
+    A reader subscribes to the FFI event queue as soon as it is handed to your
+    handler, so one that is never iterated to completion — abandoned after a
+    ``break``, or received by a handler that never reads it — must be closed
+    with :meth:`close`, or its subscription lives for the rest of the process.
+    Iterating to the end closes the reader for you.
+
+    If the stream terminates abnormally (aborted by the sender, oversized, or
+    the room disconnects mid-stream), :class:`StreamError` is raised instead of
+    a normal ``StopAsyncIteration``.
+    """
+
     def __init__(
         self,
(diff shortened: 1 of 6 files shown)


tests/rtc/test_e2e_data_streams.py
```diff
@@ -0,0 +1,487 @@
...
+Requirements:
+- LIVEKIT_URL: LiveKit server URL
+- LIVEKIT_API_KEY: API key for authentication
+- LIVEKIT_API_SECRET: API secret for authentication
+
+Tests will be skipped if these environment variables are not set.
+
+Usage:
+    pytest test_e2e_data_streams.py -v
+"""
```
1egoman, line 16: I had a LLM generate these based on the node sdk data streams v2 e2e tests. I still need to review them in depth but with a cursory glance they look good.

livekit-rtc/livekit/rtc/data_stream.py
```diff
@@ -94,48 +170,189 @@ async def read_all(self) -> str:
...
+        owns).
+        """
+        FfiClient.instance.queue.unsubscribe(self._queue)
+
+    def _close(self) -> None:
+        if not self._closed:
+            self._closed = True
+            self._unsubscribe()
+            if self._on_close is not None:
+                self._on_close()
```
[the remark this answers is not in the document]
1egoman: Fixed in 68616e3

1egoman:
I need to do another round of testing on this next week. There's been a bit more churn than I expected due to devin comments.

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -730,29 +776,43 @@ async def send_file(
         destination_identities: Optional[List[str]] = None,
         attributes: Optional[Dict[str, str]] = None,
         stream_id: str | None = None,
+        compress: bool | None = None,
```
theomonnom: We should compress by default

theomonnom (approved):

livekit-rtc/livekit/rtc/participant.py
1egoman: This reads a little weird, but to confirm, compression is enabled by default at the rust level if the flag is unset when calling into `livekit-ffi`.

I'll update this to be `compress: bool = True` here just to be unambiguous.

1egoman: FIxed in 47f6910

PR #771, opened 2026-07-30
fix(rtc): handle repeated local track subscription events

### LINKED ISSUE #770: LiveKit Python SDK - unhandled InvalidStateError in _on_room_event silently drops local_track_subscribed after a full reconnect
Hi folks, We hit a small but reproducible bug in livekit-rtc that we've had to filter out of our alerting. Still present on main today.

SDK: livekit (rtc) 1.1.8 via livekit-agents 1.6.0 · CPython 3.11.14

What happens
On a full connection restart (not a resume), the server re-sends local_track_subscribed for a LocalTrackPublication the SDK reuses. Its _first_subscription future is already resolved, so set_result(None) raises:

error running user callback for local_track_subscribed: room_handle: 5
local_track_subscribed {
  track_sid: "TR_AuA9ZzcJUN2tu"
}
Traceback (most recent call last):
  File "/app/.venv/lib/python3.11/site-packages/livekit/rtc/room.py", line 675, in _listen_task
    self._on_room_event(event.room_event)
  File "/app/.venv/lib/python3.11/site-packages/livekit/rtc/room.py", line 752, in _on_room_event
    lpublication._first_subscription.set_result(None)
asyncio.exceptions.InvalidStateError: invalid state

On main: livekit-rtc/livekit/rtc/room.py L814-818. _first_subscription is a one-shot asyncio.Future created in LocalTrackPublication.__init__ (track_publication.py L90) — no re-arm, and no .done() guard at its only call site.

Concrete example — room ******, all times UTC
• 06:57:28.653 — session close: "signal client closed: \"stream closed\"" Resume → resuming connection... attempt: 0
• 06:57:29.846 — Room reconnected — resume succeeded, no error
• 06:57:44.894 — session close: "server request to leave" ConnectionTimeout Resume
• 06:58:00.084 — resuming connection failed: wait_pc_connection timed out → restarting connection... attempt: 1
• 06:58:00.697 — Room reconnected
• 06:58:00.806 — ← InvalidStateError, 109 ms after the restart completed

Within this one room the earlier resume produced no error and only the restart path did, which lines up with the publication object being reused across a full re-subscribe.

Same pattern, second instance in the same dispatcher
room.py L867-869, track_subscription_failed branch - self._remote_participants[identity] is also unguarded, so a participant that already left raises KeyError:
Room RM_EeSEX2UcEueR, 2026-07-21 16:35:37.103 - KeyError: 'user_*******', track TR_AMegE5x4sZ2k72. That one lands 6 ms after your Rust-side could not find published track with sid for the same track (16:35:37.098), on a call where the user hung up 0.8 s after connecting.

Minor secondary ask
_listen_task reports these via a bare logging.exception(...), so they land on the root logger rather than livekit. That means we can't scope them with normal per-library log config - we ended up adding a root-logger filter matched on message text. A logging.getLogger(__name__) there would help.

swayamg20:
Fixes #770

### Problem

During a full reconnect, the server can resend `local_track_subscribed` for a reused `LocalTrackPublication`. Its `_first_subscription` future is one-shot and already resolved, so calling `set_result(None)` again raises `asyncio.InvalidStateError`. `_listen_task` catches the exception, logs an error, and drops the repeated event.

### Fix

Only resolve `_first_subscription` while it is pending. The handler still emits `local_track_subscribed` for every event, preserving callback behavior across reconnects.

The regression test dispatches the same event twice and verifies both branches:

- the first event resolves `wait_for_subscription()`;
- the repeated event does not raise and is still delivered to listeners.

### Testing

- Regression test fails with `InvalidStateError` on the unguarded handler and passes with this change.
- `MACOSX_DEPLOYMENT_TARGET=11.0 uv run pytest livekit-rtc/tests -q` — 33 passed, 14 skipped.
- Ruff lint and format checks pass across the maintained Python packages and tests.
- Mypy passes across `livekit-protocol`, `livekit-api`, and `livekit-rtc/livekit`.
- `make build-wheel PACKAGE=livekit-rtc` succeeds.

The full root test command also reached 48 passed and 30 skipped. Three audio-fixture tests could not run locally because Git LFS was unavailable and the WAV fixtures remained pointer files; CI is configured to fetch those LFS assets.

### DIFF
--- a/livekit-rtc/livekit/rtc/room.py
+++ b/livekit-rtc/livekit/rtc/room.py
@@ -814,7 +814,8 @@ class Room(EventEmitter[EventTypes]):
         elif which == "local_track_subscribed":
             sid = event.local_track_subscribed.track_sid
             lpublication = self.local_participant.track_publications[sid]
-            lpublication._first_subscription.set_result(None)
+            if not lpublication._first_subscription.done():
+                lpublication._first_subscription.set_result(None)
             self.emit("local_track_subscribed", lpublication.track)
         elif which == "track_published":
             # The participant may already have been removed by a racing
--- a/livekit-rtc/tests/test_audio_stream_room_lifecycle.py
+++ b/livekit-rtc/tests/test_audio_stream_room_lifecycle.py
@@ -550,6 +550,38 @@ def test_local_track_republished_updates_track_sid_and_repushes_metadata() -> No
     }
 
 
+@pytest.mark.asyncio
+async def test_repeated_local_track_subscribed_event_does_not_raise() -> None:
+    """A full reconnect can repeat local_track_subscribed for a publication
+    whose first-subscription future is already complete.
+    """
+    room = _make_room()
+    local = _make_local_participant("agent")
+    room._local_participant = local
+
+    track = _make_track(sid="TR_1")
+    publication = _make_local_publication(sid="TR_1")
+    publication._track = track
+    publication._first_subscription = asyncio.Future()
+    local._track_publications["TR_1"] = publication
+
+    subscribed_tracks: list[rtc.Track | None] = []
+
+    @room.on("local_track_subscribed")
+    def _on_local_track_subscribed(subscribed_track: rtc.Track | None) -> None:
+        subscribed_tracks.append(subscribed_track)
+
+    event = proto_room.RoomEvent(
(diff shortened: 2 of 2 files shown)


cloudwebrtc (approved):
lgtm

PR #779, opened 2026-08-19
fix(api): require exp claim when verifying access tokens

SashaMIT:
## Problem

`TokenVerifier.verify` decodes with issuer, HS256, and leeway, but does not require `exp`. PyJWT treats a missing `exp` as optional, so a hand-rolled token with a valid signature and no expiry verifies forever.

First-party `AccessToken.to_jwt` always sets `exp`. This matches the gap livekit/protocol#1706 closed on the Go verifier (`jwt.WithExpirationRequired`).

## Fix

When `verify_signature` is true (the production path), pass `require: ["exp"]`. `verify_signature=False` (simulate/dev) is unchanged.

Threat-model: the attacker can mint or obtain an HS256 token that omits `exp`. They do not control the verifier secret or clock. Without the claim, signature verification still succeeds and the credential never ages out. That is permanent access from a forgotten expiry, not host or agent authority.

## Test plan

- `pytest tests/api/test_access_token.py` (5/5)
- Revert-tested: removing `require: ["exp"]` makes `test_verify_token_missing_exp` fail (missing-exp token is accepted)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/livekit-api/livekit/api/access_token.py
+++ b/livekit-api/livekit/api/access_token.py
@@ -19,7 +19,7 @@ import os
 import jwt
 import dataclasses
 from dataclasses import dataclass
-from typing import Any, Optional, List, Literal
+from typing import Optional, List, Literal
 from google.protobuf.json_format import MessageToDict, ParseDict
 
 from livekit.protocol.room import RoomConfiguration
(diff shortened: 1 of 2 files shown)


lukasIO (approved):
thanks for catching this!

livekit-api/livekit/api/access_token.py
```diff
@@ -233,14 +233,28 @@ def verify(self, token: str, *, verify_signature: bool = True) -> Claims:
...
-            key=self.api_secret or "",
-            issuer=self.api_key or "",
-            algorithms=["HS256"],
-            leeway=self._leeway.total_seconds(),
-            options={"verify_signature": verify_signature},
-        )
+        # First-party minters always set exp. Without this, a hand-rolled token
+        # with a valid signature and no exp verifies forever (livekit/protocol#1706).
+        # Inline each options literal so mypy matches PyJWT's Options TypedDict
+        # (a shared dict from a ternary widens to dict[str, object]).
```
lukasIO:
```suggestion
```

we can remove this section of the comment

SashaMIT: Removed. Thanks for the review.

SashaMIT:
Dropped the mypy note. Thanks for the review.

SashaMIT:
Thanks for the review and the merge.

PR #786, opened 2026-08-21
TEL-912: Return TransferSIPParticipantResponse on TransferSIPParticipantRequests

genseric-ghiro:
## Summary
- Now returning a response for TransferSIPParticipant
- This was previously empty, even though the return type was set to `SIPParticipantInfo`, which was wrong, and resulted in systematic `SIPParticipantInfo(participant_id="", participant_identity="", room_name="", sip_call_id="")` responses.


cloudwebrtc (approved):

u9g:
The `type-check` failure here is not from your change. The submodule bump to `17c16cf` picks up `SimulationRun.List.Request` / `SimulationRun.Counts.Request`, which had a field named `from` — a Python keyword, so protoc's `--pyi_out` drops the attribute and falls back to an untyped `**kwargs`, which this repo's strict mypy rejects. Mine to fix, not yours: livekit/protocol#1738 renames the field. Once that merges, re-bump the submodule and regenerate and the two `agent_simulation.pyi` errors go away.

PR #802, opened 2026-09-02
fix(rtc): emit events to handlers in registration order

### LINKED ISSUE #778: EventEmitter dispatches handlers in Set order — peer handlers observe pre-mutation event state non-deterministically
## Summary

`livekit.rtc.EventEmitter` stores handlers in a `set` and `emit()` iterates a copy of it, so dispatch order is hash-derived rather than registration-ordered. Any handler that reads a field a peer handler mutates during the same emit gets a non-deterministic result — decided once per process, since handler identities are stable.

## Concrete case (livekit-agents 1.6.4)

`AgentActivity._on_metrics_collected` stamps `speech_id` onto LLM/TTS metrics **by mutating the emitted object** (`ev.speech_id = speech_handle.id`) before re-emitting on the session. A user handler subscribed to the same plugin's `metrics_collected` therefore sees `speech_id` either stamped or `None`, depending on which handler the set yields first.

Because the order is fixed per process, this manifests as an all-or-nothing pattern: entire sessions where 100% of plugin-level metrics are unkeyed, and other sessions at 100% keyed. Across a larger dataset it averaged out to roughly a coin flip.

## Reproduction

Minimal repro: subscribe a second handler to an LLM plugin's `metrics_collected` alongside a running `AgentSession`, record `metrics.speech_id` as observed at dispatch time, repeat over fresh processes. In 12 fresh sessions our handler observed the pre-stamp (`None`) state in 7.

## Suggestion

An insertion-ordered container for `_events` (e.g. `dict` keyed by handler, which preserves insertion order and keeps O(1) add/remove/dedup) would make dispatch deterministic and match the EventEmitter contract familiar from other ecosystems — and would let internal stamping handlers reliably run before user handlers registered later. Happy to open a PR if the direction is acceptable.

avionicharshit-byte:
fixes #778, and the approach is the one `404n0signal` proposed there.

`EventEmitter` kept handlers in a `set`, so `emit()` dispatched in hash order rather than registration order. A handler that mutates the emitted object and one that reads it therefore race, and since handler identities are stable the winner is decided once per process: whole sessions come out one way, whole sessions the other. `AgentActivity._on_metrics_collected` in livekit-agents is the live case, it stamps `speech_id` onto the metrics object before re-emitting, so a user handler on the same event sees the id or `None` for the entire run. Nothing raises and nothing logs, the metrics are just unkeyed.

`_events` now maps each event to a `dict` keyed by handler. Insertion-ordered, still O(1) add, remove and dedup, so `on` / `off` / duplicate registration behave exactly as before.

the two ordering tests use handlers with fixed hashes, so a `set` is guaranteed to yield them in the opposite order to the one they were registered in. without that they pass on main by luck: three consecutively defined functions happen to land in the set in order. on main the suite is **2 failed, 6 passed**, `At index 0 diff: None != 'speech_1'`; with the change, 8 passed.


avionicharshit-byte:
`Generating protobuf` was red , the checkout passed `ref` without `repository` so it looked for my branch here instead of the fork , that hits any fork pr so ive fixed it in a separate commit , same bug in build-protocol.yml so ive done both

@lukasIO happy to drop that ci commit if youd rather it went in on its own

avionicharshit-byte:
@cloudwebrtc could you take a look when you get a chance? rebased onto #815, green apart from the fork protobuf check.

davidzhao (approved):
lg

PR #803, opened 2026-09-03
fix failing api failover tests

lukasIO:
https://github.com/livekit/protocol/pull/1766 caused the tests to fail

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ try:
         wait_until_answered=True,
     ))
 except api.SipCallError as e:
-    print(e)                    # e.g. "SIP call failed: 486 Busy Here (resource_exhausted)"
+    print(e)                    # e.g. "SIP call failed: 486 Busy Here (failed_precondition)"
     if e.sip_status_code == 486:
         ...                     # busy
 except api.ServerError as e:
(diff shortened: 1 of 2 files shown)


stephen-derosa:
@lukasIO i think we need to revert the [upstream PR](https://github.com/livekit/protocol/pull/1766/changes) since this will be extended to a bunch of client sdks (rust included)

ChelseaBradbury:
I think this line from the readme should also be changed:
https://github.com/livekit/python-sdks/blob/3f3babb594689a07aa1ce575e3990173bbfdc67a/README.md?plain=1#L134

lukasIO:
@ChelseaBradbury can you give more context on the change and why it is required? from the API's perspective this could/should be interpreted as a breaking change?

nishadmusthafa (approved):

ChelseaBradbury:
> @ChelseaBradbury can you give more context on the change and why it is required? from the API's perspective this could/should be interpreted as a breaking change?

Basically `ResourceExhausted` isn't the right code for this, since it's likely to be interpreted as an infrastructure capacity issue rather than an expected potential call outcome. We had some discussion back and forth and we're pretty confident this isn't likely to break anything, so we decided to move forward.

ChelseaBradbury (approved):
non-blocking -> docs update

PR #806, opened 2026-09-05
add support for RPC interceptors

davidzhao:
allowing telemetry hooks so we can provide automatic traces when used in agents.

### DIFF
--- a/livekit-rtc/livekit/rtc/participant.py
+++ b/livekit-rtc/livekit/rtc/participant.py
@@ -236,6 +236,17 @@ class Participant(ABC):
 
 
 RpcHandler = Callable[["RpcInvocationData"], Union[Awaitable[Optional[str]], Optional[str]]]
+
+
+class _ChainTimeoutError(Exception):
+    """A ``TimeoutError`` raised *inside* the incoming RPC chain (by the handler or an
+    interceptor), wrapped so it is distinguishable from the response deadline expiring."""
+
+    def __init__(self, error: asyncio.TimeoutError) -> None:
+        super().__init__(str(error))
+        self.error = error
+
+
 F = TypeVar(
     "F", bound=Callable[[RpcInvocationData], Union[Awaitable[Optional[str]], Optional[str]]]
 )
@@ -633,10 +644,24 @@ class LocalParticipant(Participant):
         ``next`` counts against it; when it passes, the chain is cancelled and the caller
         gets ``RESPONSE_TIMEOUT``. Cancellation from outside (the room disconnecting) maps to
         ``RECIPIENT_DISCONNECTED``, as before.
+
+        A ``TimeoutError`` raised by the handler or an interceptor itself (an HTTP client
+        timing out, say) is not the response deadline: it propagates as an application
+        error rather than being reported to the caller as ``RESPONSE_TIMEOUT``.
         """
         handle = _chain_incoming(list(self._rpc_interceptors), self._invoke_rpc_handler)
+
+        async def _guarded() -> Optional[str]:
+            try:
+                return await handle(invocation)
+            except asyncio.TimeoutError as e:
+                # tag it so it cannot be mistaken for wait_for's own deadline expiry below
+                raise _ChainTimeoutError(e) from e
+
         try:
-            return await asyncio.wait_for(handle(invocation), timeout=invocation.response_timeout)
+            return await asyncio.wait_for(_guarded(), timeout=invocation.response_timeout)
+        except _ChainTimeoutError as e:
+            raise e.error from e.error.__cause__
         except asyncio.TimeoutError:
             raise RpcError._built_in(RpcError.ErrorCode.RESPONSE_TIMEOUT) from None
         except asyncio.CancelledError:
--- a/livekit-rtc/livekit/rtc/participant.py
+++ b/livekit-rtc/livekit/rtc/participant.py
@@ -48,11 +48,17 @@ from .track_publication import (
     TrackPublication,
 )
 from .transcription import Transcription
-from .rpc import RpcError
+from .rpc import (
+    RpcCallInfo,
+    RpcError,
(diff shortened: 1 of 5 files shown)


livekit-rtc/livekit/rtc/participant.py
```diff
@@ -595,6 +628,68 @@ async def _handle_rpc_method_invocation(
...
+        deadline = loop.call_later(invocation.response_timeout, _on_deadline)
+        try:
+            return await chain_task
+        except asyncio.CancelledError:
+            if deadline_fired:
+                raise RpcError._built_in(RpcError.ErrorCode.RESPONSE_TIMEOUT) from None
+            # cancelled from outside: awaiting propagated the cancel into the chain; let it
+            # finish unwinding before answering the caller
+            if not chain_task.done():
+                await asyncio.wait([chain_task])
```
xianshijing-lk: I know the code is also doing the unbound wait, but wonder if it makes sense to bound the wait here, like 
```
if not chain_task.done():
    await asyncio.wait([chain_task], timeout=_CANCEL_TIMEOUT)
```

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -595,6 +628,68 @@ async def _handle_rpc_method_invocation(
...
+        caller's response is built. The response deadline is enforced by the caller around
+        the whole chain.
+        """
+        handler = self._rpc_handlers.get(invocation.method)
+        if not handler:
+            raise RpcError._built_in(RpcError.ErrorCode.UNSUPPORTED_METHOD)
+
+        if asyncio.iscoroutinefunction(handler):
+            return cast(Optional[str], await handler(invocation))
+        return cast(Optional[str], handler(invocation))
```
xianshijing-lk: [nitpick](https://pullpo.io/cc?l=nitpick) 

I wonder if it is slightly cleaner to do 
```
result = handler(invocation)
if inspect.isawaitable(result):
    result = await result
return cast(Optional[str], result)
```

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -595,6 +628,68 @@ async def _handle_rpc_method_invocation(
...
+
+        A ``TimeoutError`` raised by the handler or an interceptor itself before the
+        deadline (an HTTP client timing out, say) is not the response deadline: it
+        propagates as an application error rather than being reported as
+        ``RESPONSE_TIMEOUT``.
+        """
+        handle = _chain_incoming(list(self._rpc_interceptors), self._invoke_rpc_handler)
+        loop = asyncio.get_running_loop()
+        # ensure_future: `next` continuations are typed as Awaitable, not Coroutine
+        chain_task: asyncio.Future[Optional[str]] = asyncio.ensure_future(handle(invocation))
```
xianshijing-lk, line 667: [question](https://pullpo.io/cc?l=question) 

any chance that some incoming RPCs are no-interceptor + sync handling ? that just run inline with zero scheduling ? If that is a valid use case, probably we don't want to allocate a task and timer for them ?

xianshijing-lk (approved):
some comments from a quick look, look good to me assuming you are addressing them.

livekit-rtc/livekit/rtc/participant.py
davidzhao: we would want to observe everything. so I wouldn't create an exception around observability.

rpc handlers are always async AFAIK, and I think that's the right thing to ensure no sync operations could take place here

livekit-rtc/livekit/rtc/participant.py
davidzhao: good suggestion. thx!

livekit-rtc/livekit/rtc/participant.py
davidzhao: Good call, done in 91a23d8, though it needed more than the timeout: awaiting the chain directly forwards the outer cancel into the chain and keeps this task parked until the chain finishes, so the wait here would only have started once a stubborn handler had already stopped (at the caller's deadline). The chain is now awaited through asyncio.shield, so an outside cancel is raised here at once; we then cancel the chain explicitly, wait up to 2s for it to unwind, and log a warning naming the method if it does not stop. Test added with a handler that swallows cancellation.

livekit-rtc/livekit/rtc/participant.py
davidzhao: done in 91a23d8

livekit-rtc/livekit/rtc/participant.py
```diff
@@ -595,6 +633,83 @@ async def _handle_rpc_method_invocation(
...
+            # keep this task parked until the chain finished, so a handler that ignored
+            # cancellation held up room.disconnect() for as long as the caller's deadline.
+            return await asyncio.shield(chain_task)
+        except asyncio.CancelledError:
+            if deadline_fired:
+                raise RpcError._built_in(RpcError.ErrorCode.RESPONSE_TIMEOUT) from None
+            # cancelled from outside: stop the chain and let it unwind before answering the
+            # caller, but not for long; this is the path room.disconnect() waits on
+            chain_task.cancel()
+            _, pending = await asyncio.wait([chain_task], timeout=_RPC_CANCEL_UNWIND_TIMEOUT)
```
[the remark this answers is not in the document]
davidzhao: Confirmed and fixed in 4c828f5. Reproduced it: with the shield, once our future is cancelled Python removes the shield's callback on the chain, so a RuntimeError raised from the handler's cancellation cleanup showed up as 'Task exception was never retrieved' at GC, both when the handler stopped within the unwind bound and when it outlived it. The chain's outcome is now consumed in both cases (directly after the wait, or via a done callback on the still-pending task) and logged as a warning with the traceback. Regression test captures the loop's exception handler and asserts nothing reaches it.

PR #808, opened 2026-09-08
TEL-1026: Utility for extracting transfer error details from error- #999

genseric-ghiro:
## Summary

- Adding utility for extracting transfer error details from error


livekit-api/livekit/api/twirp_client.py
```diff
@@ -109,25 +119,36 @@ def sip_status(self) -> Optional[str]:
...
-        }
+        parts = []
+        if transfer_reason is not None:
+            parts.append(transfer_reason)
+        if code is not None:
+            reason = self.metadata.get("sip_status")
+            parts.append(f"{code} {reason}" if reason else str(code))
+        what = "SIP transfer failed" if transfer_reason is not None else "SIP call failed"
+        result = f"{what}: {', '.join(parts)} ({self.code})"
+        extra = {k: v for k, v in self.metadata.items() if k not in _SIP_META_KEYS}
```
xianshijing-lk, line 150: curiously, is there any specific reason that we want to filter out sip_transfer_id from the logs here ? 

Is it intentional ?

tests/api/test_livekitapi.py
```diff
@@ -564,6 +565,43 @@ def test_sip_no_answer():
...
+
+
+def test_sip_transfer_reason():
+    err = _transfer_error(sip_transfer_reason="STR_RINGING_TIMEOUT", sip_transfer_id="STR_abc")
+    assert isinstance(err, SipCallError)
+    assert err.sip_transfer_reason == "STR_RINGING_TIMEOUT"
+    assert err.sip_transfer_id == "STR_abc"
+    # No SIP response was involved in this failure.
+    assert err.sip_status_code is None
+    assert "STR_RINGING_TIMEOUT" in str(err)
```
xianshijing-lk, line 584: curiously, should we also do 
assert "STR_abc" in str(err)

In case "STR_abc" should be in the error str ?

xianshijing-lk (approved):
two questions or nits, 

Lgtm assuming you will be addressing them

livekit-api/livekit/api/twirp_client.py
genseric-ghiro: Agree with you, it shouldn't be filtered out, added it back

tests/api/test_livekitapi.py
genseric-ghiro: Done

PR #815, opened 2026-09-14
fix(rtc): avoid deprecated asyncio.iscoroutinefunction in EventEmitter.on

### LINKED ISSUE #790: Python 3.14: EventEmitter uses deprecated asyncio.iscoroutinefunction
### Description

`livekit` 1.1.14 emits a `DeprecationWarning` on Python 3.14 whenever an RTC callback is registered because `livekit/rtc/event_emitter.py` calls `asyncio.iscoroutinefunction`. That API is deprecated in Python 3.14 and slated for removal in Python 3.16; `inspect.iscoroutinefunction` is the supported replacement.

### Reproduction

```python
from livekit.rtc.event_emitter import EventEmitter

emitter = EventEmitter()
emitter.on("event", lambda: None)
```

Run with Python 3.14 and warnings enabled:

```console
python -W default::DeprecationWarning reproduction.py
```

Observed warning:

```text
livekit/rtc/event_emitter.py:160: DeprecationWarning: asyncio.iscoroutinefunction is deprecated and slated for removal in Python 3.16; use inspect.iscoroutinefunction() instead
```

### Environment

- Python 3.14.0
- livekit 1.1.14
- Linux x86_64

### Suggested change

Import `iscoroutinefunction` from `inspect` and use it at the callback validation site. A regression test with `DeprecationWarning` promoted to an error on Python 3.14 would prevent recurrence.

xbt-a4224j:
Fixes #790

### Problem

`EventEmitter.on()` calls `asyncio.iscoroutinefunction()` to reject coroutine callbacks. That function is deprecated in Python 3.14 and slated for removal in 3.16 ([CPython gh-122858](https://github.com/python/cpython/issues/122858)), so on 3.14 every callback registration emits a `DeprecationWarning`. `Room` and friends register many handlers, so this is noisy in practice — the existing test suite alone produced 14 of them.

### Fix

`inspect.iscoroutinefunction()` is the documented replacement, but it is not a drop-in: `asyncio.iscoroutinefunction` additionally treats any callable tagged with the private sentinel `asyncio.coroutines._is_coroutine` as a coroutine function. On Python 3.9, `inspect.iscoroutinefunction(AsyncMock())` returns `False` — a bare swap would let a mocked async callback past the guard, so `emit()` would call the mock and silently discard the unawaited coroutine.

The fix keeps the sentinel check alongside `inspect.iscoroutinefunction`, read with `getattr(..., None)` so it degrades cleanly if the private attribute is removed in 3.16. Behaviour is identical to `asyncio.iscoroutinefunction` on every supported version and the deprecated function is never called.

### Testing

Four tests added to `tests/rtc/test_emitter.py`, all running on every version in the matrix:

- `test_on_does_not_warn` — promotes `DeprecationWarning` to an error around registration.
- `test_on_rejects_async_callback` — `async def` and `functools.partial`-wrapped.
- `test_on_rejects_sentinel_tagged_callback` — exercises the sentinel branch directly.
- `test_on_rejects_async_mock` — the real-world case; on 3.9 only the sentinel catches it.

Verified as a real flip against the released `livekit` wheel on all six supported Pythons (3.9–3.14): unpatched 1 failed / 7 passed / 14 warnings → patched 8 passed / 2 warnings.


cloudwebrtc (approved):
lgtm

PR #821, opened 2026-09-17
fix(rtc): filter media events from the room FFI subscription

### LINKED ISSUE #812: Room.connect FFI subscription is unfiltered; audio_stream_event fans out O(rooms × events) and starves the asyncio loop
## Summary

[PR #564](https://github.com/livekit/python-sdks/pull/564) added ``filter_fn`` to ``FfiQueue.subscribe()`` and applied it to ``AudioStream`` / ``VideoStream``. ``Room.connect()`` still does an unfiltered long-lived subscribe:

```python
self._ffi_queue = FfiClient.instance.queue.subscribe(self._loop)
```

``Room._listen_task`` only consumes ``room_event`` and ``rpc_method_invocation``, then always does ``put_nowait`` + ``await self._room_queue.join()``. Every ``audio_stream_event`` is therefore delivered to **every** connected ``Room`` and costs a full event-loop round-trip each time.

## Production impact

We run many 1:1 voice rooms in one Python process (device WebSocket gateway → ``rtc.Room`` per session). With default 10 ms audio frames:

- 8 rooms: ~7.5k listen-task wakeups/s — barely holds
- 16 rooms: ~30k wakeups/s — event loop saturates

Symptoms match #564 and #390: ``room.connect()`` futures convoy-resolve after 10–60 s, WebSocket keepalive pings time out, CPU is idle (GIL / single loop), not compute-bound.

``AudioSource.capture_frame()`` is a **short-lived** subscribe waiting on ``capture_audio_frame`` (the request callback), so it should **not** be filtered to ``audio_source_event``.

## Suggested fix

In ``Room.connect()``, subscribe with a filter matching what ``_listen_task`` actually reads:

```python
self._ffi_queue = FfiClient.instance.queue.subscribe(
    self._loop,
    filter_fn=lambda e: e.WhichOneof(message) in (room_event, rpc_method_invocation),
)
```

This is the same pattern as #564, just completing it for ``Room``. Happy to send a PR if that helps.

## Environment

- livekit==1.1.8 / livekit-api==1.1.0
- one process, many ``rtc.Room`` instances
- no ``frame_size_ms`` on ``AudioStream`` (defaults to 10 ms)

swayamg20:
## Summary

Fixes #812.

Filter `audio_stream_event` and `video_stream_event` from the long-lived subscription created by `Room.connect()`. Audio/video streams already have their own subscriptions, so forwarding these events to every room only adds event-loop callbacks and room-queue work.

Keep all other events eligible, including the `publish_track` and `unpublish_track` callbacks that `LocalParticipant` receives through the room queue. The temporary connect subscription and `AudioSource.capture_frame()` are unchanged.

Thanks to @wizd for the report and the correction identifying the publication-callback requirement. The separate publication-state robustness issue is intentionally out of scope. This change also leaves the connection-cancellation work in #813 untouched.

## Regression coverage

The new tests exercise the real room connection, FFI queue, room listener, and participant publication methods with a fake native request boundary:

- Audio/video events schedule no room callbacks while still reaching an independent media subscriber.
- Room events, RPC invocations, publication callbacks, and audio capture callbacks still arrive.
- Public publish/unpublish calls complete, and disconnect cleans up subscriptions.

Reverting only the filter makes both media regressions fail: 100 media events schedule 200 callbacks instead of the expected 100. Restoring it passes all eight new cases. This is deterministic dispatch-count evidence, not a production throughput benchmark.

## Validation

- Focused regression and existing FFI queue tests: **15 passed**, with warnings treated as errors.
- Full `tests/` and `livekit-rtc/tests/` suite: **121 passed, 72 skipped** on Python 3.13.12/macOS arm64. Live-server credentials were unset.
- Repository-wide Ruff lint and formatting: passed.
- Mypy across protocol/API/RTC packages and the new test: passed, **100 source files**.
- RTC wheel build and `git diff --check`: passed.

Local fixture note: git-lfs is unavailable in this checkout. The full-suite run used the exact upstream WAV files in a temporary directory, verified against the tracked LFS SHA-256 hashes; a pytest collection hook redirected only the three audio test modules' fixture directory. No assertions or tracked fixtures were changed. The existing JWT test-key and AudioStream test-double destructor warnings remain (34 warnings). Skipped/live-server tests have not been validated.


cloudwebrtc (approved):
lgtm

PR #826, opened 2026-09-19
rtc: dispatch event handlers in registration order

### LINKED ISSUE #778: EventEmitter dispatches handlers in Set order — peer handlers observe pre-mutation event state non-deterministically

ubmids:
Fixes #778.

`EventEmitter` stored handlers in a `set`, so `emit` walked them in hash order. That order is stable for the life of a process and arbitrary between runs, which is an awkward shape: a handler that reads a field a peer handler writes during the same emit is consistently right on one machine and consistently wrong on another, rather than flaky on both.

The reported case is in livekit-agents, where `AgentActivity._on_metrics_collected` stamps `speech_id` onto the emitted object before re-emitting it, so a second subscriber to the same plugin's `metrics_collected` sees the stamped value or `None` depending on which handler the set happens to yield first. #778 measured the pre-stamp state in 7 of 12 fresh processes.

`_events` is now a dict keyed by the handler, which is what #778 suggested. Insertion order is preserved and `setdefault` keeps the idempotent add a set gave, so registering the same handler twice still dispatches once and does not move it to the back.

Worth flagging up front:

- this is a behaviour change, though not one anything could have been relying on deliberately, since the previous order was hash-derived
- `emit` still walks a copy, so an `off` inside a handler takes effect from the next emit and not the one that removed it. `once` depends on that, and there is a test pinning it, because an ordered container is an easy place to accidentally start iterating the live collection
- scope is this emitter only. the agents-side symptom in #778 is downstream of it

On the tests: they pin handler hashes rather than registering a few callbacks and hoping a `set` reorders them. With three handlers that would pass on the broken implementation about one run in six, and a test that only usually fails is not much of a regression test. Four of the six fail on `main`; the other two cover the dedup and copy-during-emit behaviour, which were already correct and are there to stop the change quietly breaking them.

Checked with `pytest livekit-rtc/tests` (63 passed, 14 skipped), `ruff check`, `ruff format --check` and `mypy`.

Written with AI assistance; I have gone through every line and can speak to it.


davidzhao:
fixed by #802

